PR #27, opened 2021-02-07
Fix some typos in meow-cheatsheet.el

moesasji:
This just fixes some minor typos that I spotted when looking through the code that are probably worth fixing


DogLooksGood:
thank you very much!

PR #29, opened 2021-02-07
Docstring improvements and small typo fixes

moesasji:
This pull request contains the remaining set of fixes to docstrings and typos in comments. The main thing to look at carefully are the changes in meow-commands.el. In that file some of the docstrings did not apply to the specific command. I think I fixed them correctly though. 

The only one that isn't clear to me from the docstring is what the universal argument is supposed to do in meow-tail, meow-head, so please double-check carefully that the suggested fix is indeed correct.


DogLooksGood:
Awesome! @moesasji 

Thank you very much!

PR #34, opened 2021-02-14
Fix a wrong direction in a docstring

moesasji:
Title says it all; I spotted a docstring that wasn't fully correct. This commit fixes that.


DogLooksGood:
Thanks!

PR #36, opened 2021-02-16
a few small changes

Makima0:
`delete-selection-mode` default is `-1`, I think user settings should be respected

put wdired-mode into insert mode


DogLooksGood:
Hi, First of all, thanks for contribution!

I just have a better tweak (IMO) for wdired. It can not only switch to NORMAL state when you enter wdired, but also can switch back to MOTION state when you exit wdired.

DogLooksGood:
solved, please check.

PR #55, opened 2021-06-18
some simple instructions on KMACRO in both English and Chinese

Dicridon:
Added some simple instructions and an example about how to use kmacros in meow. A Chinese version tutorial is also added.


DogLooksGood:
感谢！
一点小建议，我觉得中文 kmacro 文档的链接，可以放在中文的 readme 中。

DogLooksGood:
Thanks!

PR #81, opened 2021-12-02
Add meow-describe-thing-at-point command

krdzo:
Here is a "`describe-at-point`" function. It's "inspired" from evils `evil-lookup` command that is usually bound to `K` in evil.
It has `meow-describe-thing-at-point-funciton` variable where you can change which function is called for description lookup. For now it is bound to `woman`, like in evil, I couldn't find a better built-in command, but something like `elisp-slime-nav-describe-elisp-thing-at-point` from [elisp-slime-nav package](https://github.com/purcell/elisp-slime-nav) would be the perfect default option for this.
I tried to do something like:
```
(defcustom meow-describe-thing-at-point-funciton
  #'(lambda () (call-interactively #'describe-symbol))
  "Function that used by `meow-describe-thing-at-point'."
  :group 'meow
  :type 'function)
```
but that looked ugly so I didn't include it.

### DIFF
--- a/COMMANDS.org
+++ b/COMMANDS.org
@@ -283,3 +283,7 @@ Swap secondary selection with current selection.
 *** meow-pop-grab
 
 Pop secondary selection.
+
+*** meow-describe-thing-at-point
+
+Look up for help for thing at point
--- a/meow-command.el
+++ b/meow-command.el
@@ -1527,6 +1527,16 @@ This command is a replacement for build-in `kmacro-end-macro'."
           (region-str (buffer-substring-no-properties rbeg rend)))
      (meow--second-sel-set-string region-str))))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; HELP LOOKUP
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defun meow-describe-thing-at-point ()
+  "Look up help for thing at point.
+Calls `meow-describe-thing-at-point-funciton'"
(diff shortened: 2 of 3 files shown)


DogLooksGood:
Hi, first of all, thanks for the contribution.

But I'm afraid I have a few questions here. Since Meow is designed to integrate tightly with Emacs defaults. I'd like to have to a describe command baked by some Emacs __built-in__ variables. Since there's no 3rd packages will care about our `meow-describe-thing-at-point-function`, so users must update this variables based on their need.

So my suggestion is not to add this function, let user bind whatever they like. For example, if user always wants a `describe-symbol`, they can do
```
(meow-normal-define-key '("K" . describe-symbol))
```

If their `K` won't behave differently based on major mode, they can do
```
(meow-normal-define-key '("K" . "C-x M-k"))

(define-key emacs-lisp-mode-map (kbd "C-x M-k") #'describe-symbol)
(define-key shell-mode-map (kbd "C-x M-k") #'woman)
```
It looks a bit complex, but add `meow-describe-thing-at-point-function` won't make it simple, because no other packages will maintain this variable for us. If there's an built-in variable, that every package will take care, it will be a great idea.

Sorry for dispointing you.
Feel free to ask any question, if my explaination doesn't make sense to you.

krdzo:
Aha. No problem. I just setup something like this for my setup and thought to generalize it and that it would be something useful for this package, but I didn't understand the intention behind this package and didn't ask if it would be good idea to implement this. 
As you can see I too didn't like that there isn't a better built-in option for `meow-describe-thing-at-point-function`, like you said something that every major-mode can set it to according to it needs, so it looked to me that this is a fine mechanism (to let the user manage the variable).
It didn't occur to me that you can achieve the same effect with this "define-key/kbd macros" method.

All in all than you for your answer and sorry if I wasted you time on this.

DogLooksGood:
Yes, more works need to be done, to help people understanding the package. 

Thank you.

PR #103, opened 2021-12-07
Add meow-excluded-modes to disable meow-mode in specified major-mode

naoking158:
Thank you for developing and maintaining the awesome package!

I would like to have the ability to disable meow-mode in specified major-modes where meow-mode is not needed, such as Eshell-mode.

### DIFF
--- a/meow-core.el
+++ b/meow-core.el
@@ -97,7 +97,9 @@ This minor mode is used by meow-global-mode, should not be enabled directly."
 ;;;###autoload
 (define-global-minor-mode meow-global-mode meow-mode
   (lambda ()
-    (unless (minibufferp)
+    (if (or (minibufferp)
+            (memq major-mode meow-excluded-modes))
+        (meow-mode 0)
       (meow-mode 1)))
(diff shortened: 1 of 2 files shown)


DogLooksGood:
First of all, thanks for the contribution!

But there are some choices are made by design. One of them is do not disable `meow-mode` in a specific mode.
I believe you feel inconvenient when open a new eshell buffer, because you have to switch to insert state. But if you disable `meow-mode` for Eshell mode, you no longer have a consistent leader globally. That will require you to distinguish which type of buffer you are using, for example if you switch to a Eshell buffer with something like `SPC x o`, you can not go back with the same keybinding.

To solve this problem, for using meow with all kinds of shells and repls, what I suggest is:
```
(add-hook 'eshell-mode-hook 'meow-insert)
```
This will give you the insert state for the first time you open a shell.

How do you think?

naoking158:
I'm sorry for the late reply.

I understood your thought, and your suggestion will solve everything!

Thank you for your help, and I close this pull request.

Issue #112, opened 2021-12-09
Org citations and deletion/change


mclearc:
I've noticed that in org-mode `meow-delete` and `meow-change` don't work as expected inside citation brackets. Say I've got:

`[cite:@beiser1987 chap. 2; @franks2000; @forster2012a; @digiovanni2020]`

placing the cursor after the first bracket and calling `meow-delete` kills the entire line, which is unexpected. Similarly with `meow-change`. Using either of these commands elsewhere within the bracketed citations also deletes the rest of the line. Is this expected behavior? It seems like a bug, but I'm new to meow so... 🤷‍♂️

mclearc:
Yeah --- it seems problematic that this is tied to a keybinding rather than to the actual function, given that different modes will modify this keybinding in different ways (e.g. the function apparently being called in org was `citar-org-delete-citation`). This is going to generate surprises for people. Is there a way to make it so that it just calls `delete-char`?

DogLooksGood:
You can,
- give `C-d` a command to actually delete a char, that's most major modes will do
- or particularly for this case, `(setq meow--kbd-delete-char "<deletechar>")`.

`<deletechar>` is your `Delete` key on keyboard.

Issue #109, opened 2021-12-08
meow-undo comments line instead of undoing


anhsirk0:
meow version: 20211207
Layout: Qwerty Layout
I downloaded meow via package-install

pressing 'u' or M-x meow-undo doesnot undo instead it comments current line

skmd0:
Can you post your meow config here?

anhsirk0:
;; Meow setup
(defun meow-setup ()
  ;; (setq meow-cheatsheet-layout meow-cheatsheet-layout-qwerty)
  
  (meow-motion-overwrite-define-key
   '("j" . meow-next)
   '("k" . meow-prev))
  (meow-leader-define-key
   ;; SPC j/k will run the original command in MOTION state.
   '("j" . "H-j")
   '("k" . "H-k")
   ;; Use SPC (0-9) for digit arguments.
   '("1" . meow-digit-argument)
   '("2" . meow-digit-argument)
   '("3" . meow-digit-argument)
   '("4" . meow-digit-argument)
   '("5" . meow-digit-argument)
   '("6" . meow-digit-argument)
   '("7" . meow-digit-argument)
   '("8" . meow-digit-argument)
   '("9" . meow-digit-argument)
   '("0" . meow-digit-argument)
   '("/" . meow-keypad-describe-key)
   '("?" . meow-cheatsheet))
  (meow-normal-define-key
   '("0" . meow-expand-0)
   '("9" . meow-expand-9)
   '("8" . meow-expand-8)
   '("7" . meow-expand-7)
   '("6" . meow-expand-6)
   '("5" . meow-expand-5)
   '("4" . meow-expand-4)
   '("3" . meow-expand-3)
   '("2" . meow-expand-2)
   '("1" . meow-expand-1)
   '("-" . negative-argument)
   '(";" . meow-reverse)
   '("," . meow-inner-of-thing)
   '("." . meow-bounds-of-thing)
   '("[" . meow-beginning-of-thing)
   '("]" . meow-end-of-thing)
   '("a" . meow-append)
   '("A" . meow-open-below)
   '("b" . meow-back-word)
   '("B" . meow-back-symbol)
   '("c" . meow-change)
   '("d" . meow-delete)
   '("D" . meow-backward-delete)
   '("e" . meow-next-word)
   '("E" . meow-next-symbol)
   '("f" . meow-find)
   '("g" . meow-cancel-selection)
   '("G" . meow-grab)
   '("h" . meow-left)
   '("H" . meow-left-expand)
[...]

PR #113, opened 2021-12-09
FAQ document

skmd0:
I have created a new document to address the common question that are posted in issue tracker (#109, #112).

I am not sure how exactly the solution should be done, so please feel free to change anything.

If this is not what you had in mind we can discuss what specifically we should do instead and close this PR.


DogLooksGood:
looks good to me, let's merge it!

PR #119, opened 2021-12-10
make meow-beacon supress keymaps as well

Not-Leader:
Prevent self-insert keymaps from shadowing meow's beacon mode map


DogLooksGood:
Thank you!!

PR #121, opened 2021-12-10
Give priority to meow's beacon map

Not-Leader:
Adding suppress-keymap didn't fully prevent meow's beacon keymap from
being shadowed, so i had to do this as well


DogLooksGood:
This should be added.

Issue #117, opened 2021-12-09
Intermittent issues with keypad-mode


fedreg-bn:
Just discovered keypad-mode and it's great!!!

Mine seems to not always work as expected though..

Sometimes it pops up for a split second, disappears right away, and has an error:
```
Failed Select: Invalid argument
```
This happens if I push `SPC-c`, `SPC-x`, or `SPC-c` and with various major modes.

But it is inconsistent and sometimes works fine..

video demo:
https://user-images.githubusercontent.com/87655715/145484571-fd36efb6-cbfa-4451-ab8c-8582746e157f.mov

Most likely on my end but not sure where to look...

Keep a pretty minimal config with only these packages
```
(ag
cider
company
fzf
key-chord
magit
meow
rainbow-delimiters
smartparens
typit
xclip)
```
Emacs 28.0.50
Mac Big Sur
Running in TTY mode in terminal

If anyone has experienced / resolved this issue please let me know.   I'll try to dig around in meow's code this weekend and see if I can spot the conflict

Thanks!

DogLooksGood:
Hi, 
Try `M-x toggle-debug-on-error`, then when something goes wrong, you'll get a stacktrace.

That information will be helpful.

fedreg-bn:
Ah, good idea! 

```
Debugger entered--Lisp error: (file-error "Failed select" "Invalid argument")
  read-event(nil t 2305843009213693951)
  sit-for(2305843009213693951 t)
  meow-describe-keymap((keymap (120 . kmacro-to-register) (118 . kmacro-view-macro-repeat) (116 . kmacro-swap-ring) (115 . kmacro-start-macro) (114 . apply-macro-to-region-lines) (113 . kbd-macro-query) (112 . kmacro-cycle-ring-previous) (110 . kmacro-cycle-ring-next) (108 . kmacro-call-ring-2nd-repeat) (107 . kmacro-end-or-call-macro-repeat) (105 . kmacro-insert-counter) (102 . kmacro-set-format) (101 . kmacro-edit-macro-repeat) (100 . kmacro-delete-ring-head) (99 . kmacro-set-counter) (98 . kmacro-bind-to-key) (97 . kmacro-add-counter) (81 . kdb-macro-redisplay)))
  meow--keypad-display-message()
  meow--keypad-try-execute()
  meow-keypad-self-insert()
  funcall-interactively(meow-keypad-self-insert)
  call-interactively(meow-keypad-self-insert nil nil)
[...]

PR #123, opened 2021-12-11
Add a keymap close delay variable and replace fixnum

Fedreg:
References issue #117 

Not sure if this is useful to anyone else but is working great for me.

I was having issues with `sit-for` blowing up [HERE](https://github.com/emacs-mirror/emacs/blob/3af9e84ff59811734dcbb5d55e04e1fdb7051e77/lisp/subr.el#L2711).

Merge if anyone else has mentioned this issue.  Just ignore otherwise.   

Happy to change any naming, add to changelog, etc.. 

Thanks!!

### DIFF
--- a/CUSTOMIZATIONS.org
+++ b/CUSTOMIZATIONS.org
@@ -203,7 +203,7 @@ Default: ~0.5~
 
 The delay in seconds before popup keybinding descriptions appear.
 
-** meow-keypad-describe-open-delay
+** meow-keypad-describe-close-delay
 
 Default: ~2000000~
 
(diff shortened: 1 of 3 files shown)


DogLooksGood:
Hi, thanks for the contribution.

The close delay is not necessary to be configurable, we just give `sit-for` a large number to make sure it will last "forever" in real cases.  Thus, there's no need to add new variable.  Please undo the variable changes.

Fedreg:
@DogLooksGood Got it.   No prob.  Thought you might not want a hard coded number there so added the var.  But removed all that and just kept an int instead.   

Hadn't seen your last comment on issue #117 saying to do just that, apologies.

So this is the only change in the PR now:
image

I can squash these commits down or open a new PR with just the original change if you prefer.   Thanks!

DogLooksGood:
squash and merged. (you don't have to squash on your side)
Thank you!

Fedreg:
@DogLooksGood Thanks for the help!!

PR #124, opened 2021-12-11
add keyboard cheat sheets and mention tutor to TUTORIAL.ORG

skmd0:
Meow was quickly mentioned in the [latest System Crafters stream](https://youtu.be/7E-6j6VTZgM?t=1122)! David (from System Crafters) after reading the description, he clicked on Tutorial link under Documents section. I got the feeling that section is a bit misleading because people expect to learn more about keybindings as well.

I think we should at very top of Tutorial section add links to all the keyboard layout cheatsheets. So people who skipped 'Getting started' section (because they we curious how Meow works before installing) can still get the information they need and not get discouraged from using Meow.

What do you think?

### DIFF
--- a/TUTORIAL.org
+++ b/TUTORIAL.org
@@ -1,9 +1,23 @@
 #+title: TUTORIAL
 
-Meow is a modal editing with 5 states: NORMAL, INSERT, MOTION, KEYPAD and BEACON.
+* Default Keybindings Cheat Sheet
+
+Check out default keybindings for your keyboard layout:
+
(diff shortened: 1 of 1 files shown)


DogLooksGood:
IMO, the word "default" is not accurate, since we do not make those keybindings builtin.
How about change the word "default" to "recommended" or something else?

How do you think?

skmd0:
Good point! Will change it. How about I just change the heading to Keybindings Cheat Sheet and right below that I add this:

>We have created some recommended configs for specific keyboard layouts that you can re-use for quick setup.

PR #125, opened 2021-12-11
Add suggestions for doc and mewo-var

Linerre:
- meow-kill doc suggestions
- Set normal mode as the default entering mode for gud-mode (GNU debugger)
- re-order meow-mode-state-list alphabetically: a-z A-Z


Linerre:
meow-kill 那里我好像忘把这句加回来了：
>This command support fallback behaviour, default to `~meow-C-k~`.

PR #126, opened 2021-12-12
add mode and meow state specific keymaps

turbana:
I would like to propose adding meow state specific and mode specific keybinds. So keybinds that would only be active under a specified mode (major or minor) and when under a specific meow state. This would allow binds in any combination of mode and meow state to be active and play nicely with one another.

My thought was to use `emulation-mode-map-alists` (similar to meow state's minor modes) to accomplish this. For lack of a better name I chose to call these keymaps "overlay maps", let me copy and paste some documentation going over their implementation:

```
;;; Overlay Maps
;; An "overlay map" is a keymap that should be activated when running under a
;; specific mode (major or minor) and meow state. A keymap is considered an
;; overlay when the vector [meow-overlay] is set as a key in the map. These
;; overlay maps are stored in the mode's keymap under the key
;; [meow-X-state] where X is the meow state. Ex. [meow-insert-state].
;;
;; During setup we add the symbol `meow--user-overlay-maps' to
;; `emulation-mode-map-alists' to facilitate our overlay maps.
;;
;; When meow changes state the following occurs:
;; 1. we look through all active keymaps for any overlay maps.
;; 2. all such overlay maps are combined into a single keymap with
;;    `make-composed-keymap'.
;; 3. we ensure the combined map is accessible under `emulation-mode-map-alists'
;;    by adding a mapping of meow minor mode to keymap to the alist
;;    `meow--user-overlay-maps'.
```

An example:
```
(meow-state-define-key 'org-mode-map 'normal
  "S-<left>"    'org-promote-subtree
  "S-<up>"      'org-move-subtree-up
  "S-<down>"    'org-move-subtree-down
  "S-<right>"   'org-demote-subtree
  "C-S-<left>"  'org-do-promote
  "C-S-<right>" 'org-do-demote
  )
```

Thoughts?

### DIFF
--- a/CUSTOMIZATIONS.org
+++ b/CUSTOMIZATIONS.org
@@ -17,6 +17,18 @@ Define key in NORMAL state.
    ...)
 #+end_src
 
+** meow-state-define-key
+
+Bind =keys= under =keymap= when running in meow =state=. Both =keymap= and =state= can be either a quoted list or single quoted item.
+
+The keys bound here are found through =emulation-mode-map-alists= and so have a higher precedence than the specified =keymap=. See the elisp manual on [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Searching-Keymaps.html][Searching Keymaps]] for more information on keymap order.
+
+Example:
+#+begin_src emacs-lisp
+  (meow-state-define-key 'emacs-lisp-mode-map 'insert
+    '("C-z" . eval-last-sexp))
+#+end_src
+
 ** meow-leader-define-key
 
 Like ~meow-normal-define-key~, but for leader keymap.
--- a/meow-core.el
+++ b/meow-core.el
@@ -27,6 +27,7 @@
 
 (require 'meow-util)
 (require 'meow-command)
+(require 'meow-keymap)
 (require 'meow-keypad)
 (require 'meow-var)
 (require 'meow-esc)
@@ -172,7 +173,9 @@ then SPC will be bound to LEADER."
   (when (meow--init-motion-p)
     (meow-normal-mode -1)
     (meow--save-origin-commands)
-    (meow-motion-mode 1)))
+    (meow-motion-mode 1))
+  ;; ensure user overlay keymaps are initialized
+  (meow-activate-overlay-maps))
 
 (defun meow--disable ()
   "Disable Meow."
@@ -203,6 +206,9 @@ then SPC will be bound to LEADER."
                        `((meow-keypad-mode . ,meow-keypad-state-keymap)))
   (add-to-ordered-list 'emulation-mode-map-alists
                        `((meow-beacon-mode . ,meow-beacon-state-keymap)))
+  (add-to-ordered-list 'emulation-mode-map-alists
+                       'meow--user-overlay-maps)
+  (add-hook 'meow-switch-state-hook #'meow-activate-overlay-maps)
   (when meow-use-cursor-position-hack
     (setq redisplay-highlight-region-function #'meow--redisplay-highlight-region-function)
     (setq redisplay-unhighlight-region-function #'meow--redisplay-unhighlight-region-function))
@@ -221,6 +227,10 @@ then SPC will be bound to LEADER."
   (remove-hook 'kill-emacs-hook 'meow--on-exit)
   (meow--disable-shims)
   (meow--remove-modeline-indicator)
+  ;; remove meow overlay keymaps
+  (setq emulation-mode-map-alists
+        (delq 'meow--user-overlay-maps emulation-mode-map-alists))
+  (remove-hook 'meow-switch-state-hook #'meow-activate-overlay-maps)
   (when meow-use-cursor-position-hack
     (setq redisplay-highlight-region-function meow--backup-redisplay-highlight-region-function)
     (setq redisplay-unhighlight-region-function meow--backup-redisplay-unhighlight-region-function))
--- a/meow-helpers.el
+++ b/meow-helpers.el
@@ -22,6 +22,7 @@
 ;; Define custom keys in normal map with function `meow-normal-define-key'.
 ;; Define custom keys in global leader map with function `meow-leader-define-key'.
 ;; Define custom keys in leader map for specific mode with function `meow-leader-define-mode-key'.
+;; Define custom keys for specific meow state and specific mode `meow-state-define-key'.
 
 ;;; Code:
 
@@ -66,6 +67,26 @@ Optional argument ARGS key definitions."
         args)
   (cl-loop for arg in args do
            (add-to-list 'meow--motion-overwrite-keys (car arg))))
+
+(defun meow-state-define-key (keymaps states &rest args)
+  "Define keys under KEYMAPS for STATES. KEYMAPS can either be a quoted keymap
+or a list of quoted keymaps. STATES can either be a quoted meow state or a
+list of quoted meow states.
+
+Usage:
+  (meow-state-define-key 'org-mode-map 'insert
+    '(\"TAB\" . company-expand))
+  (meow-state-define-key '(prog-mode-map org-mode-map) '(normal insert)
+    '(\"Q\" . keyboard-quit))"
+  (declare (indent 2))
+  (dolist (state (if (listp states) states (list states)))
+    (dolist (keymap-sym (if (listp keymaps) keymaps (list keymaps)))
+      (let* ((keymap (symbol-value keymap-sym))
+             (overlay-map (meow--get-overlay-map keymap state t)))
+        (dolist (key-def args)
+          (define-key overlay-map
+            (kbd (car key-def))
+            (meow--parse-def (cdr key-def))))))))
 
 (defun meow-setup-line-number ()
   (add-hook 'display-line-numbers-mode-hook #'meow--toggle-relative-line-number)
--- a/meow-keymap.el
+++ b/meow-keymap.el
@@ -22,6 +22,9 @@
 
 ;;; Code:
 
+(require 'cl-lib)
+(require 'subr-x)
+
 (require 'meow-var)
 
 (defvar-local meow--origin-commands nil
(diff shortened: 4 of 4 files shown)


DogLooksGood:
Thanks for the contribution, it seems to be both well implemented and well documented.

It's clear to me that a lot users want this feature.  Before accepting the PR, I want to have a explanation here.

In an early version of Meow, it is possible to define a mode specific leader keymap in meow, at that moment I believe the leader keymap is for highly used commands. And which commands are highly used depends on current major mode. But I removed the functionality later because I realized that it's not a good idea to have custom keybindings __only existing__ in meow's keymap. I want meow to be something that leverage your vanilla Emacs configuration, not the package that limits how you build your configuration.

That is:
- meow must not be invasive
- meow's configuration can be and should be limited in a single `.el` file, just like other packages
- remove meow from your configuration should be as simple as deleting a single file

Additionally, these cases are what I'm trying to avoid:
- configure meow for every major/minor modes
- removing meow = giving up the effort you spend to organize your keybinding
- introducing meow = rebuilding your keybind system

So my recommendation is to bind keys in vanilla Emacs.  For example:

You want a consistent keybinding <kbd>SPC t i</kbd> to toggle images in both org-mode and markdown-mode. You should make a consistent keybinding in vanilla Emacs, and tell meow to use that.

```emacs-lisp
;; assume C-x M-t the prefix for 'toggle'
;; global 'toggle' commands
(global-set-key (kbd "C-x M-t l") 'display-line-numbers-mode)
;; major mode specific 'toggle' commands
(define-key markdown-mode-map (kbd "C-x M-t i") #'markdown-toggle-inline-images)
(define-key org-mode-map (kbd "C-x M-t i") #'org-toggle-inline-images)
;; add entry for 'toggle' in leader
(meow-leader-define-key '("t" . "C-x M-t"))
```

Then you have consistent keybinding <kbd>SPC t i</kbd> in leader and <kbd>C-x M-t i</kbd> in vanilla Emacs.

Below configuration are not specific to meow, they are useful whether you use meow or not.

---

I'd like to have a discussion here.  If I can convince others, I'm still inclined to reject this feature.

turbana:
Thanks for the compliment and the detailed reply!

You've made a compelling argument for sure. The keybind system is still one of the parts of emacs that I can never quite get fully sunk into by brain. I'm coming from the evil-mode side of emacs (and vim before that) and so was originally trying to replicate the way you can bind keys there (and with general.el being able to specific evil `:state`). And so I hadn't even really considered having a common prefix key in bare emacs and binding a key in meow to that prefix. One of my goals was to have a local leader key that was mode specific and that neatly solves that problem without further code and in bare emacs. So thanks for that!

The other problem I'm looking to solve that the above doesn't work for is to have a command called under a specific emacs mode AND specific meow state. For example when in `org-mode` I want the `>` key to attempt to expand a `yasnippet` and if not found enter a bare `>`. I would like that to only occur in insert state and leave all other states open. In evil/general.el I have the following:

```
;; By convention I title most of my snippets {snippet}> so I would like the
;; > key to try and expand snippets automatically. When no snippet is found
;; a regular > should be entered.
(defun ic/yas-expand-> ()
  (interactive)
  (insert ">")
  (yas-expand))

(general-define-key
 :states 'insert
 :keymaps 'org-mode-map
 ">" 'ic/yas-expand->)
```

That would be one situation where having a mode and state specific keymap would be handy.

That being said I don't have very many insert state specific commands. Most are for normal state which can be handled in the manner you pointed out. So for those few commands I do want state specific functionality I suppose I can just inspect the meow minor mode. So the above could be rewritten as:

```
(defun ic/yas-expand-> ()
  (interactive)
  (cond ((bound-and-true-p meow-insert-mode)
         (insert ">")
         (yas-expand))
        ((bound-and-true-p meow-normal-mode)
         ;; normal mode code
         )
        (t
         ;; other mode code, etc
         )))
```

But I'm happy with only doing that for the few commands I feel need to be state aware and using your method for the bulk of my other keybinds. I can see where having the option to have meow state specific keybinds will lead people to tie their keybinds to meow and make it harder to remove it. I'm certainly finding that to be the case for evil-mode!

So thanks again for the nudge in the right direction for organizing my keybinds "the emacs way". I'm happy to close the PR and look forward to whats next for meow!

DogLooksGood:
I also have some insert mode specific keybindings. I think my case is very similar to your case mentioned here.

In clojure-mode, I want the semicolon key to insert a colon when the cursor is not at the 1) beginning of line 2) string 3) comment. Otherwise it should still insert the semicolon.

In rust-mode, I want to the minus key to insert a underscore when the cursor is after a word.

[here is how I implement this in my configuration](https://github.com/DogLooksGood/dogEmacs/blob/master/elisp/init-rust.el#L65)

I just bind the command in clojure-mode-map and rust-mode-map. Meow normal state won't be affected in this case.

In fact, if you look at my configuration, you can find `meow` in just four places. 
- `init-modal.el` the meow's configuration
- `init-modeline.el` my custom mode line format
- `init-rime.el` use meow's predication to disable input method in non-insert state
- A list of themes, for some faces used by meow

turbana:
> I just bind the command in clojure-mode-map and rust-mode-map. Meow normal state won't be affected in this case.

Thanks for this! It gave me the "ah ha" moment realizing that meow's normal state keymap is part of `emulation-mode-map-alist` and therefore takes precedence over both local modes and major modes (where insert specific commands can go).

This is where I have trouble fully groking the way emacs looks up keybinds, but it's slowly starting to sink in.

ChauhanT:
Hi, not sure if this is the right place to ask. I am trying to remap `<` in org-mode to `'org-shiftleft`. I am trying to do this with general with something like : 
```
(general-define-key
 :keymaps 'org-mode-map
 "<" 'org-shiftleft)
```

Sadly, this does not work as it also maps the `<` in the insert mode :( Has anyone been able to tackle this ?

DogLooksGood:
This is not supported.  The normal state is a general state, there's no mode-specific normal state keymap.  

Here the reason and solution is explained: https://github.com/meow-edit/meow/pull/126#issuecomment-992004368

rickalex21:
@DogLooksGood  Hello, I would like to switch to meow but I would like to switch back and forth
between meow and evil until I move everything over to meow.

I was reading your suggestion to use vanilla emacs to define keys but I'm not sure
how to account for normal, insert, and motion. I'm farily new to meow I don't want to
have two configs.

I'm not sure if a macro is the best way to go about this but it should:

* Switch between evil and meow by setting `my/modal`.
* Define keys in normal, insert,visual, and motion.
* Define keys in major modes.
* Use leader key, I use space.

Ideally I would like to do an `&rest rest`parameter so I won't have to write
my/define-key on every line, evil can handle the `&rest` but I would have to
figure that out for meow.

This is would be great:

```emacs-lisp
(setq my/modal 'meow)

(my/define-key 'normal
               (kbd "<leader>z z") (lambda ()(interactive) (message "hi"))
               (kbd "<leader>z f") #'do-something)
```

I suppose I could use `(concat "<leader>" ,key)` for evil. How does meow know if the
key is for normal or insert?  If I'm using the leader key, normal mode would be assumed? This is what I
came up with:

```emacs-lisp
(setq my/modal 'evil)

(defmacro my/define-key (state map &rest rest)

(pcase my/modal
 ('evil `(evil-define-key ,state ,map ,rest)) 
 ('meow  `(meow-leader-define-key '(,key . ,func ) ))
 (_ (display-warning :wrong-variable "Wrong variable set for my/modal"))))


(my/define-key 'normal global-map  "zz" (lambda () (interactive)(message "hi zz")) )
```

The current problem I have is getting arguments out of a list then passing them to evil
or meow. The macro will do this which is invalid.

```emacs-lisp
(evil-define-key 'normal global-map ("zz" (lambda nil (interactive) (message "hi zz"))))
```

Would like this, then figure out how to implement it for meow.

```emacs-lisp
(evil-define-key 'normal global-map "zz" (lambda nil (interactive) (message "hi zz")))
```
Thanks

DogLooksGood:
@rickalex21 I think I completely got your idea, but it sounds to be against the concept of this package. 

Meow has only one leader, it doesn't matter it's in normal mode or motion mode. Use `meow-leader-define-key` to create a keybinding in current leader map without "<leader>", but it's not recommended to have a complex leader system in meow since it's not reusable, and not accessible in insert state. I guess this package will be really weird when you are trying to duplicate the evil's experience.

Personally, I have few leader key bindings. All rest are default, and invoked by keypad.
https://github.com/DogLooksGood/meomacs/blob/master/private_template.org#leader

rickalex21:
It looks like I will create a bunch of  `c-c` and to use them with meow leader. That's the awesome thing about meow that you can reuse existing keys. Will have to look at that thanks.

licht1stein:
How do I define this: I want `j` and `l` in `dired-mode` to go in and out of directories in normal mode?

licht1stein:
Makes me feel stupid :) thank you!

zackattackz:
Hi I do a similar thing as [here](https://github.com/meow-edit/meow/pull/126#issuecomment-992004368) where I bind mode-specific key maps to be different functions than whats set in the global map. Everything works fine, the function I set gets run instead of the global one. However in the key pad the display name is still the global function, not the mode-local one that is actually going to be run. This is minor but was wondering if there was a fix?

DogLooksGood:
@zackattackz Can you show a snippet about what you have done?

zackattackz:
Yes so I have the following for bindings on `C-b`

```
(global-unset-key (kbd "C-b"))
(global-set-key (kbd "C-b b") #'ivy-switch-buffer)
(global-set-key (kbd "C-b q") #'kill-current-buffer)
(global-set-key (kbd "C-b n") #'next-buffer)
(global-set-key (kbd "C-b p") #'previous-buffer)
(define-key git-commit-mode-map (kbd "C-b b") #'run-term ; run-term is just here to test for now
```

Then in my `meow-setup` I have:

```
(setq meow-cheatsheet-layout meow-cheatsheet-layout-qwerty)
(meow-leader-define-key
  '("b" . "C-b")
```

On my scratch buffer pressing `C-b` gives:
image

And `C-b b` runs `ivy-switch-buffer`

Inside the git-commit-mode, pressing `C-b` gives:
image

Although it says `C-b b` is bound to `ivy-switch-buffer`, pressing `b` here will actually run `run-term` as intended.

zackattackz:
There is more in my configs that could possibly be affecting this but I shared only what I think is relevant. But if it can't be reproduced let me know and I can share full init.el. I don't expect you to go out of your way to solve my problem but I very much appreciate meow and I hope I can contribute back some day.

zackattackz:
Ok thanks for the advice with using C-c, makes sense will use it going forward.

I attempted your suggestion and it seems that actually pressing `C-c b b` will have the intended affect based on buffer.

However using the meow keypad (`<SPC> b b`) will only ever display the global bindings and names, and it will only ever run the globally bound function.

I will paste what I did below in case it's an error on my end.

Thanks

```
(defvar my-buffer-keymap
  (let ((keymap (make-keymap)))
	(define-key keymap "b" #'ivy-switch-buffer)
	(define-key keymap "n" #'next-buffer)
	(define-key keymap "p" #'previous-buffer)
	(define-key keymap "q" #'kill-current-buffer)
    keymap))
(defalias 'my-buffer-keymap my-buffer-keymap)
(global-set-key (kbd "C-c b") 'my-buffer-keymap)

(defvar my-git-commit-mode-keymap
  (let ((keymap (copy-keymap my-buffer-keymap)))
    (define-key keymap "b" #'run-term)
    keymap))
(defalias 'my-git-commit-mode-keymap my-git-commit-mode-keymap)
(define-key git-commit-mode-map (kbd "C-c b") 'my-git-commit-mode-keymap)
```
I cleared any related entries in `meow-leader-define-key`

DogLooksGood:
@zackattackz Guess you are missing `(setq meow-keypad-leader-dispatch "C-c")`?

zackattackz:
Yup, good call.

Adding that fixes the functionality of the keypad so that the right command is running now, my bad.

But the keypad still reads "b -> ivy-switch-buffer" rather than run-term.

zackattackz:
Interestingly the keypad does properly switch between "my-buffer-keymap" and "my-git-commit-mode-keymap" on the top level though

zackattackz:
Maybe I am going about using meow wrongly though. My original intention is to try to have consistent binds across different modes, as well as for the keypad to guide me in case I forgot the exact key mappings. Similarly to how spacemacs works, but I didn't want to use spacemacs itself or evil mode.

Should I go about this a different way?

DogLooksGood:
I think it's a display bug, just had a fix with f6a1b48eb5678b225666b8726962d26a81588d8e

zackattackz:
Thank you, works great!

A7R7:
I recently found a workaround to bind mode and meow-state specific keybinds, without hacking into meow. You may check [this discussion](https://github.com/meow-edit/meow/discussions/523)

Issue #71, opened 2021-11-05
Is it possible to add a description to prefix keys in keypad?


bsag:
First, thanks for writing meow! I'm really enjoying how powerful and streamlined it is.

I've created a few top-level items using `meow-leader-define-key`, some of which I've built with a letter prefix to group similar commands together, e.g.:

```
   '("w w" . ace-window)
   '("w u" . winner-undo)
   '("w r" . winner-redo)
   '("w h" . split-window-below)
   '("w v" . split-window-right)
   '("w c" . delete-window)
   '("w o" . delete-other-windows)
```

This works well, enabling me to press `SPC w u` for winner undo for example, but at the top level menu, the 'w' prefix is just labelled `+prefix`. 

Is there some way I am missing to provide a more descriptive label like "windows"? I know that you can do that with which-key, but I'm not sure how to do that when meow is providing the items.

Thanks for any help you can provide!

DogLooksGood:
Here's an example if you are using KEYPAD mode.

```
(defvar window-keymap
  (let ((keymap (make-keymap)))
    (define-key keymap (kbd "w") #'ace-window)
    (define-key keymap (kbd "u") #'winner-undo)
    (define-key keymap (kbd "r") #'winner-redo)
    (define-key keymap (kbd "h") #'split-window-below)
    (define-key keymap (kbd "v") #'split-window-right)
    (define-key keymap (kbd "c") #'delete-window)
    (define-key keymap (kbd "o") #'delete-other-windows)
    keymap))

;; define an alias for your keymap
(defalias 'window-keymap window-keymap)

(global-set-key (kbd "C-x C-w") 'window-keymap)
;;                              ^ note the quote
```

Then in KEYPAD popup, it will be displayed as `+window-keymap`.

I think the alias way is the built-in way in Emacs, [how kmacro-keymap is defined in emacs source code](https://github.com/emacs-mirror/emacs/blob/3af9e84ff59811734dcbb5d55e04e1fdb7051e77/lisp/kmacro.el#L203).

---

But if you define your keys in leader, there's no key description. The popup description in Meow is only for KEYPAD. The reason is leader is designed to be simple, since the shortcuts in leader is not baked by a normal key binding, it is not recommended to have complex keymap in leader.

If you want `SPC w ?`, you should define `C-w ?`, and add `'("w" . meow-keypad-start)` in leader.

---

[...]

PR #131, opened 2021-12-16
Helper function to create prefix keys for leader commands

bitspook:
I am kinda new to Emacs lisp, and I had a hard time figuring out how to add new prefixes to the leader key. `(error "Key sequence z s starts with non-prefix key z")` didn't help much, but enough that I could find my way to write a helper function.

I think it can be useful for others as well. Please take a look. If you think it can be an acceptable addition to the repo and want me to make some changes, I'll be very happy to do that.

Thank you for your work on this project.

### DIFF
--- a/meow-helpers.el
+++ b/meow-helpers.el
@@ -31,6 +31,31 @@
 (require 'meow-var)
 (require 'meow-keymap)
 
+(defmacro meow--define-keymap-prefix (parent-keymap key name)
+  "Create prefix KEY for PARENT-KEYMAP with NAME."
+  (let ((keymap-name (make-symbol (concat name "-keymap")))
+        (alias-name (make-symbol name)))
+    `(progn
+       (defvar ,keymap-name (make-sparse-keymap))
+       (defalias ',alias-name ,keymap-name)
+       (define-key ,parent-keymap (kbd ,key) ',alias-name))))
+
+(defun meow-leader-define-prefix (key name)
+  "Define key prefix for leader.
+
+Defines KEY as a prefix which can be used in `meow-leader-define-key`.
+NAME is a valid symbol name which is also displayed in which-key popup.
+
+Usage:
+    (meow-leader-define-prefix \"g\" \"my-git\")
+
+    (meow-leader-define-key
+     '(\"gs\" . magit-status)
+     '(\"gb\" . magit-blame)
+     '(\"gg\" . magit-dispatch))
+"
+  (eval (macroexpand `(meow--define-keymap-prefix meow-leader-keymap ,key ,name))))
+
 (defun meow-leader-define-key (&rest args)
   "Define key for Leader.
 


DogLooksGood:
Hi, thanks for the contribution!

It seems like `meow-leader-define-prefix` is added for define a prefix, which is useful for `which-key` and keypad state.
It had been asked in https://github.com/meow-edit/meow/issues/71. Additionally, users may care about building a nested leader key system. Personally, I'm against this idea and I have a detailed explanation in this PR: https://github.com/meow-edit/meow/pull/126. 

My thought is we just create a keymap, alias it, then have a simple binding in meow. Users may not be aware of this solution, but that is just a problem for documentation.

How do you think?

bitspook:
What you suggest makes perfect sense. I'll close my PR and see if I can find a good place in documentation to add this. Thank you for taking time to respond. 

I suppose this is how I'd accomplish what you're suggesting:

```lisp
(defvar spook-git-keymap (make-sparse-keymap)
  "My custom keymap for easily accessing git stuff")
(defalias 'spook-git spook-git-keymap)

(define-key spook-git-keymap (kbd "s") #'magit-status)

(global-set-key (kbd "C-c g") 'spook-git)

(meow-leader-define-key
 '("g" . "C-c g"))
```

Small problem here is, even though on doing `C-c` which-key can show `+spook-git`, but `Spc` shows `meow-dispath-NN`

meow-git-help

Is it possible to somehow get a good help text for `SPC` as well in this case?

bitspook:
I figure I can just do 

```lisp
    (meow-leader-define-key
     '("p" . projectile-command-map))
```

DogLooksGood:
Yes, you can bind a key to a keymap. 

If you bind like this
```
(meow-leader-define-key
 '("g" . "C-c g"))
```
The command/keymap for `g` is not determined until it is called.
The remapping by other packages can happen at any time that meow is not aware of.

bitspook:
I've found that if I do the following, sometimes the keybindings clash(?) and don't do what I expect/want.

```lisp
(meow-leader-define-key
 '("g" . "C-c g"))
```

E.g "C-c g g", assigned to `magit-dispatch` works if I call it with `C-c g g`, but if I do `SPC g`, it don't show up in the listing, and pressing `g` anyway shows this:

image

So I figured I'll just add keymaps directly. When I add `projectile-command-map`, it works as expected; however, using my custom keymaps give me an error saying: `command-execute: Wrong type argument: commandp, spook-windows-keymap`. 
As far as I can tell, projectile is defining the keymap just like I am.

```lisp
    (defvar spook-buffer-keymap (make-sparse-keymap))
    (defalias 'spook-buffer spook-buffer-keymap)

    (global-set-key (kbd "C-c b") spook-buffer-keymap)

    (define-key spook-buffer-keymap (kbd "b") #'ivy-switch-buffer)
    (define-key spook-buffer-keymap (kbd "d") #'kill-current-buffer)
    (define-key spook-buffer-keymap (kbd "s") #'startup--get-buffer-create-scratch)
```

bitspook:
I figured it out. I have to use the alias i.e `spook-buffer` and not the keymap itself to have it work with meow. I don't understand why, but I am kinda tired of fumbling since morning, so I am gonna let this one pass.

Here's how I am configuring my stuff now.

I've create a macro to help me:

```lisp
  (defmacro spook-defkeymap (name prefix &rest bindings)
    "Create a new NAME-keymap bound to PREFIX, with BINDINGS.

  Usage:
    (spook-defkeymap \"spook-git\" \"C-c g\"
      '(\"s\" . magit-status))
  "
    (let* ((keymap-name (intern (concat name "-keymap")))
           (keymap-alias (intern name))
           (keymap-bindings (mapcar
                             (lambda (binding)
                               (let ((binding (eval binding)))
                                 `(define-key ,keymap-name (kbd ,(car binding)) #',(cdr binding))))
                             bindings)))
      `(progn
         (defvar ,keymap-name (make-sparse-keymap))
         (defalias ',keymap-alias ,keymap-name)
         (global-set-key (kbd ,prefix) ',keymap-alias)
         ,@keymap-bindings)))
```

I define my keymap like this:

```lisp
    (spook-defkeymap
     "spook-buffers" "C-c b"
     '("b" . ivy-switch-buffer)
     '("d" . kill-current-buffer))
```

Then assign them to meow-leader-key

```lisp
    (meow-leader-define-key
     '("p" . projectile-command-map)
     '("w" . spook-windows)
     '("b" . spook-buffers)
     '("g" . spook-git))
```

I get nice help in popups:
image

and keys don't clash either.

Sorry for the spam, I thought I should put it out here in case someone else might stumble here.

DogLooksGood:
Firstly, `g` is for `C-M-`, since there's no `C-g` in the key seq, so using `g` here is safe.  Keypad has a "fallback to literal key" feature.  That is, you can use `SPC c a` for `C-c a`, if there's no keybinding on `C-c C-a`. Since `g` is always available for `C-M-`, so this fallback behavior will not work in this case.

I'm just typing, and found you figured it out. :)

PR #144, opened 2021-12-22
Improve wording and punctuation in meow-tutor

yusi1:
I noticed some wording errors in meow-tutor, fixing these will make it flow better and make it easier to understand.

I love meow BTW, it is amazing!!


DogLooksGood:
Thank you very much!

PR #153, opened 2021-12-25
fix meow--enable to avoid executing symbol-name w.r.t lambda func.

naoking158:
In `meow--enable`, it checks whether "self-insert" is assigned to key "a" or not by using `symbol-name`.
However, an error occurs when key "a" is assigned to a lambda function, such as neotree.
To avoid this problem, the assigned function is checked for symbol before `symbol-name` is executed.


DogLooksGood:
thank you!

PR #154, opened 2021-12-26
Fix beacon symbol returned in meow--current-state

eshrh:
If `meow-beacon-mode` is active, then `meow--current-state` will return the symbol `'cursor`. `meow--switch-state` expects the symbol to be `'beacon`. This may be a bug, but I could be wrong since everything seems to work fine.


DogLooksGood:
It's a bug, thank you for correcting this!

PR #155, opened 2021-12-26
Allow user to define custom states

eshrh:
This patch allows the user to easily define custom states. Currently, a lot of functionality is hard-coded around the 5 core states in meow. This makes creating a new state somewhat inconvenient, so I've also rewritten and changed several functions.

I think that the ability to define new states is pretty essential to a flexible modal editing system. Evil has `(evil-define-state)`, and ryo-modal, modalka, etc are basically just defining states. It opens up many possibilities, for instance I have a custom state for smartparens commands like moving by sexps and slurping/barfing. Its useful for packages that introduce too many keybinds to bind to free keys (like smartparens/paredit). 

My smartparens mode before this patch: https://gist.github.com/eshrh/f01f605902e51df4bdb22b4a77df90c0
After this patch: https://gist.github.com/eshrh/94bba3effd3bb4d15faf5f332d9fa421

Please see the commit messages and docstrings for detailed information

I realize this is a significant change, and so I would like some feedback before putting in time to polish the details.


DogLooksGood:
Hi, this could be a huge change, but it looks good to me. 

Personally, I have no interest in adding new states to meow at the moment.  A state in meow has to address a problem that must be solved in modal editing layer.  If something can be achieved with a simple transient keymap, that does not fall under meow's responsibility.

But, of course, I'm happy to see this improvements.  It will definately introduce a better abstraction to this package.  If you decide to go head with this, I'll be happy to help when you run into any problems.

eshrh:
Thanks! I've polished some of the macros, added some docstrings, and wrote an entry in CUSTOMIZATIONS.org

DogLooksGood:
Thanks! 

Essentially looks good, just a few minor thoughts:
1. `meow-keymap-alist` use string state names, I'd like change it to symbol. If we consider state name a constant, it should always be symbol.
2. Avoid `eval`, they seems to be unnecessary.  If you know a list of symbols of states, you can call whatever functions you defined.
3. Avoid storing code block in `meow-update-cursor-functions-alist`, using function symbol is good in this case I believe.
4. We can add `:interactive nil` to `define-minor-mode` to avoid the minor mode be called as command. 
5. Should we consider making the minor mode function can be used directly, instead of providing a `xx-mode-enter`? 

I'll also give try with current implementation, to make sure everything works well before we merging it.

eshrh:
I'll change 1-3. For 4 and 5:

Currently, `meow--switch-state` is the last function that I couldn't rewrite elegantly. It's because meow-motion-mode has extra conditions. The big problem is that `meow--update-indicator` and `meow--update-cursor` are called from `meow--switch-state` only, and not from each minor mode -- this is why it should not be interactive now. 

However, it should be easy to make each minor mode call the cursor/indicator updates inside the init function. I think it's very natural as a user to call `meow-normal-mode` directly. Should I rewrite all the init functions and move those functions out of `meow--switch-state`?

Personally i think this is good, but I didn't make the change earlier because I didn't want to change existing functionality. 

This would make the switch-state function mostly useless, but i think it's fine to keep the function (rewrite to use the mode list) so that all the commands and shims don't also need to be changed.

DogLooksGood:
For the reason I point out 4-5 is that when we merge  into master, the documentation is also updated. I'd like to use `meow-xxx-mode` instead of a enter function.

Do you have any idea bout how motion init can be improved?a

eshrh:
Alright, i'll try to implement it. I don't know how to improve the motion init, because saving the origin command seems necessary. However, the normal state disabling is redundant because we do that anyway in the init.

For 2. can you show me where I can avoid eval? My issue is that I can't map a macro like (bound-and-true-p) over a list, since it takes the arguments literally, I need to unquote it. Is there a better way?

eshrh:
Ok, I think I've fixed all the problems you mentioned!

DogLooksGood:
https://github.com/meow-edit/meow/pull/155/commits/28907fd1cd6ab15afa12efcdb324867b0ad8ef5a#diff-350669cfe6e43ace91995b81e7b427acc6fc1ac46d691727fd676fe73e6a803cR132

If normal state is enabled, we can not get the original command, because it is override by commands in normal state.

PR #158, opened 2021-12-29
Allow user to define custom states

DogLooksGood:
Test a bit before merging into master.


DogLooksGood:
it seems like sometimes can not exit beacon state when cursor leaving the secondary selection.

DogLooksGood:
Current `meow-define-state` can not be used to define existing states in meow. I'd like to make some changes make it possible.

fbergroth:
> Current `meow-define-state` can not be used to define existing states in meow. I'd like to have some changes to make it possible.

Maybe `meow-generate-define-key` can be used to define `meow-motion-overwrite-define-key` by removing `meow--motion-overwrite-keys`, something like this:
```diff
diff --git a/meow-helpers.el b/meow-helpers.el
index 0a01add..019338d 100644
--- a/meow-helpers.el
+++ b/meow-helpers.el
@@ -63,9 +63,7 @@ Optional argument ARGS key definitions."
           (define-key meow-motion-state-keymap
             (kbd (car key-def))
             (meow--parse-def (cdr key-def))))
-        keybinds)
-  (cl-loop for keybind in keybinds do
-           (add-to-list 'meow--motion-overwrite-keys (car keybind))))
+        keybinds))
 
 (defun meow-setup-line-number ()
   (add-hook 'display-line-numbers-mode-hook #'meow--toggle-relative-line-number)
diff --git a/meow-util.el b/meow-util.el
index c272b1b..509e1e1 100644
--- a/meow-util.el
+++ b/meow-util.el
@@ -450,9 +450,10 @@ For performance reasons, we save current cursor type to
 
 (defun meow--save-origin-commands ()
   (setq meow--origin-commands nil)
-  (cl-loop for key in meow--motion-overwrite-keys do
+  (cl-loop for key-code being the key-codes of meow-motion-state-keymap do
            (ignore-errors
-             (let ((cmd (key-binding (kbd key))))
+             (let* ((key (char-to-string key-code))
+                    (cmd (key-binding (kbd key))))
                (when (and (commandp cmd)
                           (not (equal cmd 'undefined)))
                  (let ((rebind-key (concat meow-motion-remap-prefix key)))
diff --git a/meow-var.el b/meow-var.el
index ab66569..429bbb8 100644
--- a/meow-var.el
+++ b/meow-var.el
@@ -454,10 +454,6 @@ Has a structure of (sel-type point mark).")
 (defvar meow--keypad-help nil
   "If keypad in help mode.")
 
-(defvar meow--motion-overwrite-keys
-  '("SPC")
-  "A list of keybindings to overwrite in MOTION state.")
-
 (defvar meow--beacon-backup-hl-line
   nil
   "Whether hl-line is enabled by user.")
```

DogLooksGood:
@fbergroth Looks good. I believe @eshrh is making some progress for improving this PR.

But we should still keep the function definition. When user jump to the source, it's better to see a function than a macro invocation.

fbergroth:
Using `meow-generate-define-key` for every mode seems a bit excessive. Maybe just a single function is needed?
```elisp
(defun meow-bind (mode keybinds)
  "Define KEYBINDS in MODE."
  (let ((map (alist-get mode meow-keymap-alist)))
    (pcase-dolist (`(key . def) keybinds)
      (define-key map (kbd key) (meow--parse-def def)))))
```
and use it like
```elisp
(meow-bind 'leader
           '(("j" . "H-j")
             ("k" . "H-k")))
```

eshrh:
I had thought about doing this, but an earlier commit on the the branch used the macro to actually define meow-normal-define-key, so it had to be generalized if I didn't want to introduce an extremely breaking change. 

Now that the built in define-key functions are explicitly defined, I think this is a more elegant solution. This also gets me thinking about whether it's worthwhile to introduce another function that can set cursor types properly instead of having to generate a new variable.

I'm still a little bit unsure of whether to introduce two different ways to do the same thing (defining a normal state keybind) is good though. 

Would you mind writing a pull request with your change to the define-state branch?

fbergroth:
> Would you mind writing a pull request with your change to the define-state branch?

@eshrh sure, have a look at #166

> I'm still a little bit unsure of whether to introduce two different ways to do the same thing (defining a normal state keybind) is good though.

Right. I think it would be preferable to use `meow-bind` everywhere, and to declare the old functions deprecated, which isn't a breaking change.

alexluigit:
Hi @DogLooksGood.
It seems that this PR breaks `meow-motion-origin-command`.
For some reason `meow--save-origin-commands` failed to capture `SPC`. It should return something like '("SPC" . "H-SPC") if I'm not mistaken.

DogLooksGood:
@alexluigit Hi, this variable is no longer needed. 

Is that you are using this variable in your configuration?

alexluigit:
I'm sorry but I don't know which variable you are referring to (since both `meow-motion-origin-command` and `meow--save-origin-commands` are functions). 

In meow-keymap.el we have: 

```emacs-lisp
(defvar meow-leader-keymap
  (let ((keymap (make-sparse-keymap)))
    (suppress-keymap keymap t)
    (define-key keymap (kbd "SPC") 'meow-motion-origin-command)
    ;; ...
))
```

So `meow-motion-origin-command` is the default keybinding for `SPC` key in `meow-leader-keymap`, user should expect key chord `SPC SPC` to execute command that is bound to `SPC` key in that major-mode. 

As an example, in Dired `SPC SPC` should invoke `dired-next-file` command since `SPC` is bound to that command in Dired mode (by default). It worked as expected prior to this commit. Now `SPC SPC` do nothing in Dired mode. Can you verify this? Thanks!

DogLooksGood:
@alexluigit You are right, let me have a fix.

DogLooksGood:
@alexluigit Fixed.

alexluigit:
Thanks for the quick fix! It now works as expected.

PR #166, opened 2022-01-07
Meow bind

fbergroth:
See https://github.com/meow-edit/meow/pull/158#issuecomment-1007072673. Some comments:

- Removal of `meow--motion-overwrite-keys` should be tested a bit more, I only have the basic qwerty config which seems ok
- Maybe `meow-keymap-alist` should directly map to values and not symbols, and we can get rid of `symbol-value` in `meow-bind`
- I prefer not using `&rest keybinds` - much less quoting
- Documentation needs to be updated if we proceed with this

### DIFF
--- a/meow-helpers.el
+++ b/meow-helpers.el
@@ -97,9 +97,7 @@ Check `meow-normal-define-key' for usages."
           (define-key meow-motion-state-keymap
             (kbd (car key-def))
             (meow--parse-def (cdr key-def))))
-        keybinds)
-  (cl-loop for keybind in keybinds do
-           (add-to-list 'meow--motion-overwrite-keys (car keybind))))
+        keybinds))
 
 (defun meow-setup-line-number ()
   (add-hook 'display-line-numbers-mode-hook #'meow--toggle-relative-line-number)
--- a/meow-helpers.el
+++ b/meow-helpers.el
@@ -38,24 +38,28 @@ the string NAME, and the string SUFFIX"
   (intern (concat (if prefix prefix "meow") (if two-dashes "--" "-")
                   name suffix)))
 
-;; Macro to produce define-key function helpers
-;; for each of the items in the alist meow-keymap-alist.
-(defmacro meow-generate-define-key (mode keymap)
-  (let ((mode-string (symbol-name mode))
-        (keybind (gensym)))
-    `(defun ,(meow-intern-string mode-string
-                                 "-define-key") (&rest ,keybind)
-       ;; Documentation string
-       ,(concat "Define key for " mode-string " keymap \n\n"
-                "Usage: \n"
-                " (meow-" mode-string "-define-key) \n"
-                "    '(\"@\" . hs-toggle-hiding))\n"
-                "Optional argument ARGS key definitions")
-       (mapcar (lambda (key-def)
-                 (define-key ,keymap
-                   (kbd (car key-def))
-                   (meow--parse-def (cdr key-def))))
-               ,keybind))))
+(defun meow-define-keys (state &rest keybinds)
+  "Define KEYBINDS in STATE.
+
+Example usage:
+  (meow-define-keys
+    ;; state
+    'normal
+
+    ;; bind to a command
+    '(\"a\" . meow-append)
+
+    ;; bind to a keymap
+    (cons \"x\" ctl-x-map)
+
+    ;; bind to a keybinding which holds a keymap
+    '(\"c\" . \"C-c\")
+
+    ;; bind to a keybinding which holds a command
+    '(\"q\" . \"C-x C-q\"))"
+  (let ((map (alist-get state meow-keymap-alist)))
+    (pcase-dolist (`(,key . ,def) keybinds)
+      (define-key map (kbd key) (meow--parse-def def)))))
 
 (defun meow-normal-define-key (&rest keybinds)
   "Define key for NORMAL state with KEYBINDS.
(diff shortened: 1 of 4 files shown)


eshrh:
Thanks! a39d062 is a great change. As for the other two, here are some thoughts:
1. What do you think about changing the name of meow-bind? Perhaps meow-define-key?
2. I discussed this briefly with tianshu, but maybe removing &rest and making the keybinds one big list is not the best idea.

`meow-normal-define-key` and `meow-leader-define-key` are intuitive and their function is immediately obvious. I agree that quoting each command is a bit awkward, but there are some downsides. Defining normal keybinds and leader keybindings are two different actions, and so they are currently done with two calls to two different functions. I think (again, very subjectively) that this is cleaner than two calls to the same function with two different arguments. Configs should be as atomic and intuitive as possible. Many people use copied (meow-setup) functions anyway, so not many people are going to be typing a lot of quotes.

It's a super small difference, which is why I don't think it's worth obsoleting the functions. Obsoleting the meow-x-define-key functions is practically the only way to go if &rest is removed from meow-bind, since otherwise there would be two different syntaxes between the functions. 

Coming from this, maybe the best way to go is to introduce meow-bind with &rest quoted arguments, and then refactor all the define-key functions to use meow-bind underneath. So it's basically like meow-bind is the low-level way to do it, and the rest are just helper utilities.

What do you think about this?

fbergroth:
> 1. What do you think about changing the name of meow-bind? Perhaps meow-define-key?

Sure, I don't mind! `bind` was taken since it's common in `use-package`. `meow-define-key` sounds like it defines a single key, but we allow multiple.

> 2. I discussed this briefly with tianshu, but maybe removing &rest and making the keybinds one big list is not the best idea.
> 
>
> `meow-normal-define-key` and `meow-leader-define-key` are intuitive and their function is immediately obvious. I agree that quoting each command is a bit awkward, but there are some downsides.


Regarding quoting, I understand that I create some friction by not using `&rest`. My reasons for doing it was that I was a bit turned off from seeing the configuration examples of meow. It doesn't feel idiomatic, and I thought that if it was to ever be changed, now would be a good time.

I had hoped to be able to configure meow keybindings by just using `bind-key` from `use-package`, that's why I looked into this. Maybe I should look further into it, I don't remember why I stopped.

Ofcourse I can just do ``(apply #'meow-bind 'leader `(...))`` locally and be happy, but I think meow configurations would look better with the proposed change.

> Defining normal keybinds and leader keybindings are two different actions, and so they are currently done with two calls to two different functions. I think (again, very subjectively) that this is cleaner than two calls to the same function with two different arguments. Configs should be as atomic and intuitive as possible. 

I'm not sure. It's just `(meow-define-key 'leader ...)` vs `(meow-leader-define-key ...)`. Having identical implementations for every `meow-*-define-key` doesn't look nice to me. One standard way vs several ad-hoc. In most configs, all keybindings are done via the same interface.

> Many people use copied (meow-setup) functions anyway, so not many people are going to be typing a lot of quotes (on the other hand, every single user will eventually have to change their config with this change)

The deprecated functions does not have to be removed, they can be kept around indefinitely.

> It's a super small difference, which is why I don't think it's worth obsoleting the functions. Obsoleting the meow-x-define-key functions is practically the only way to go if &rest is removed from meow-bind, since otherwise there would be two different syntaxes between the functions.
> 
> Coming from this, maybe the best way to go is to introduce meow-bind with &rest quoted arguments, and then refactor all the define-key functions to use meow-bind underneath. So it's basically like meow-bind is the low-level way to do it, and the rest are just helper utilities.
> 
> What do you think about this?

I'm happy to go either way. I understand that my changes lacked a bit of context, so I tried to better motivate them above.

fbergroth:
> ... and then refactor all the define-key functions to use meow-bind underneath. So it's basically like meow-bind is the low-level way to do it, and the rest are just helper utilities.

(I did change `meow-*-define-key` to call `meow-bind` underneath)

DogLooksGood:
I'm happy to have some discussion here.

I'm not against introducing breaking changes, but I'd only like to have a breaking change when we can finally switch to vanilla `define-key`. I'm against the idea of featureless wrappers, those introducing something new, but just another way to express the same. Instead of how many characters to type, I care more about the abstractions. I don't think there's much differences when comparing following ways.
```
(define-key meow-normal-state-keymap (kbd "a") 'meow-append)

;; and in `meow-normal-define-key`

'("a" . meow-append)

;; or

("a" . meow-append)

;; or even 

"a" meow-append
```
If we are express the same thing, the syntax is not that much important, IMO.

I introducing a helper because we have to support this
```
("a" . "H-a")
```

fbergroth:
I forgot to mention that I also thought it would be nice if meow provided the basic keybindings for atleast qwerty, perhaps as an alist, and users only had to configure leader keys and any changes to default configuration.

@DogLooksGood I went with &rest in my last commit. I played around and extracted `meow-dispatch` from `meow--parse-def`, enabling one to do:
```
(define-key meow-leader-state-keymap (kbd "j") (meow-dispatch "H-j"))
```
That won't work with `bind-key` though, and I think using only `define-key` without any sugar is a bit ugly (with so many bindings), so I think it makes sense to keep `meow-define-keys`.

DogLooksGood:
Sure, meow doesn't provide default key bindings for qwerty, because I'm not satisfied with current qwerty key binding.

fbergroth:
I squashed some commits and added a fix for:

> Maybe `meow-keymap-alist` should directly map to values and not symbols, and we can get rid of `symbol-value` in `meow-bind`

DogLooksGood:
Nice! Shall I invite you to meow-edit group?

fbergroth:
> Nice! Shall I invite you to meow-edit group?

Sure :-)

DogLooksGood:
Thank you! Invitation sent.

PR #171, opened 2022-01-08
Fix a typo & propose one more candidate for meow-mode-state-list

Linerre:
I feel the need to edit *.diff or *.patch files so I suggest binding
NORMAL to diff-mode. But I notice MOTION is the default for
ediff-mode, so change it to MOTION if you think all diff files should
go with MOTION.


DogLooksGood:
Thank you!

PR #193, opened 2022-01-24
Add a customizable meow-goto-line-function

fbergroth:
I added this to be able to do: `(setq meow-goto-line-function 'consult-goto-line)`.

I also noticed `meow-goto-line` has two bindings in KEYBINDING_QWERTY.org:
```
(meow-normal-define-key
   ...
   '("Q" . meow-goto-line)
   '("X" . meow-goto-line)
```
Do you plan to keep both?

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -930,22 +930,12 @@ numeric, repeat times.
   (interactive "p")
   (meow-line n t))
 
-(defun meow-goto-line (arg)
+(defun meow-goto-line ()
   "Goto line, recenter and select that line."
-  (interactive "P")
-  (let (beg
-        end
-        (ln (when arg (prefix-numeric-value arg))))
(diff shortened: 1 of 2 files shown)


DogLooksGood:
> Do you plan to keep both?

No, I think I will keep the `X` in examples.

velppa:
This PR removed the default implementation, isn't it? I believe it should be brought back and if the user wants to use custom, she can do it via `(setq meow-goto-line-function ...)`.

fbergroth:
I believe the default implementation was covered by `goto-line`. Was it not?

DogLooksGood:
I think the default behaviour is covered by `goto-line`.

velppa:
Now I see it, stupid me, sorry for confusion.

PR #209, opened 2022-03-21
Allow better customization of meow-keypad-start-keys

eshrh:
Currently, there is no good way to have say, `u` in leader correspond to `C-c`. 

In this patch, meow-keypad-start-keys becomes an alist of keys to keys instead of a list. This is of course a breaking change, but I don't think that it will affect many people since the variable isn't actually documented. 

In general, I think this is a bit of a hack to avoid the real issue which is that keypad code is not very abstracted (meta and ctrl key variables need to be set individually, and so on) but I don't think it's worth potentially introducing some bigger more breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Master (Unreleased)
 
+### Breaking Changes
+
+* [#209](https://github.com/meow-edit/meow/pull/209) Make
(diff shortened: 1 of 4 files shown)


DogLooksGood:
Thank you! I agree with your opinion about this change. Please include an updated changelog.

PR #212, opened 2022-03-28
remove prefix args from save/kill

eshrh:
I noticed that `_arg` does not seem to be doing anything in commands that save or kill. Please merge if this is safe to remove.


DogLooksGood:
Thanks!

PR #213, opened 2022-03-29
add meow-duplicate and duplicate-and-comment

eshrh:
I have written a rudimentary pair of duplication functions that basically save and then yank. 

This is basically like `crux-duplicate-current-line-or-region` from crux, but implemented with meow functions instead. I have found this very useful after I wrote it for myself.

The current duplicate function duplicates a char instead of a line like crux when no region is active. I think this makes sense because selecting a line is natural and common but selecting a char is more awkward (expand right).

I think this is a good candidate for `P` in suggested keybindings which is currently free [not in commit (yet)]

Any thoughts on this?

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -212,6 +212,28 @@ This command supports `meow-selection-command-fallback'."
   (when (meow--allow-modify-p)
     (meow--execute-kbd-macro meow--kbd-yank-pop)))
 
+(defun meow-duplicate ()
+  "Duplicate region if active. Otherwise duplicate char at point"
+  (interactive)
+  (if (region-active-p)
+      (progn
+        (meow-save)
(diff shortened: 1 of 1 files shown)


DogLooksGood:
Hey, sorry, I'm against this proposal. 

There are some reasons:
1. This is a normal DWIM command, and meow is against this. The command should be separated as for single char, and region.
2. This command affects the kill-ring, which is unnecessary.
3. `meow-duplicate-and-comment` is too complex for its functionality.

I don't think meow should cover all the operations. Meow itself is build with operations around a selection concept, some feature will indeed be missing here, but that's acceptable.

eshrh:
ok, sounds good. this is fairly trivial to write for a personal config anyway.

PR #216, opened 2022-04-15
force reverse hint display after meow-reverse

eshrh:
Consider the following:

```

aaaa
aaaa
aaaa
|
aaaa
aaaa
```
where | is the cursor, on a blank new line. There is no good way to select the new line AND the previous three lines, except 
1. moving back lines to the first `aaaa`, then `meow-line` then 3.
2. moving up one line, then `meow-line` then 1 then `meow-reverse` then 2

None of these are nice. Compared to selecting the blank line and the NEXT two lines is just meow-line + 2.

You should be able to press `meow-line` and then immediately `meow-reverse` then 3, but `meow-reverse` doesn't work here. The reason is that reversing only exchanges point and mark, but if both are in the same place then `meow--direction-backward-p` says that we are actually going forward.


DogLooksGood:
A bit busy recently, later I'll look into this.

eshrh:
I guess I never really considered that, but https://github.com/meow-edit/meow/blob/master/meow-command.el#L911, direct backward selection on the empty line moves back to the previous line. 

I changed it so that direct backward selection does not move back and selects the blank line.

eshrh:
any update on this @DogLooksGood ?

DogLooksGood:
With this change, you can't reverse a backward selection.

Do `back-word`, then `reverse`.

eshrh:
What do you mean by do back-word first? In meow-reverse?

I fixed the issue my just not always forcing the hints to be reversed. Let me know if there was a better way

DogLooksGood:
Let me try it later.

DogLooksGood:
No, it doesn't work for me.
```
aaaa
aaaa
aaaa
|
aaaa
aaaa
```

In this case, do `meow-line`, then `reverse`. Nothing happens.
And `negative-argument`, `meow-line`, `meow-line`. The expanding doesn't respect the direction.
I think I can work on this when I have time.

fbergroth:
I think 627724198342889b62671da4ee3f36f3bf192665 from #272 also solves this case.
One case that should be added is to set `meow--backward-p` to nil whenever the selection type is changed.

eshrh:
alright, I'll try to fix it once and for all tomorrow

fbergroth:
You may also have to use `use-region-p` instead of `region-active-p` in `meow--direction-backward-p`.

eshrh:
Originally i believed that using a stateful variable to keep track of direction was bad, but now I think i believe that it is a necessity.

Using the relative position of the point and mark as the ONLY source of direction leads to edge cases like this one and will probably continue to be a bit of a problem. I think i'll go with your solution, and scrap this idea.

PR #217, opened 2022-04-17
Any state as initial state (meow-mode-state-list)

Xervon:
In some modes I want to start in insert state (shell-mode, eshell-mode, ...). Instead of adding just insert-mode this allows insert-mode and any custom state to be defined as the initial state in a given mode.

### DIFF
--- a/meow-core.el
+++ b/meow-core.el
@@ -138,22 +138,19 @@ Note: When this function is called, NORMAL state is already enabled.
 NORMAL state is enabled globally when `meow-global-mode' is used.
 because in `fundamental-mode', there's no chance for meow to call
 an init function."
-  (let ((state (meow--mode-get-state)))
+  (let ((state (meow--mode-get-state))
+        (motion (lambda ()
+                  (meow--disable-current-state)
+                  (meow--save-origin-commands)
+                  (meow-motion-mode 1))))
     (cond
      ;; if MOTION is specified
      ((eq state 'motion)
-      (meow-normal-mode -1)
-      (setq meow--current-state nil)
-      (meow--save-origin-commands)
-      (meow-motion-mode 1))
-
-     ;; if NORMAL is specified
-     ((eq state 'normal)
-      nil)
+      (funcall motion))
 
      (state
-      (meow-normal-mode -1)
-      (funcall (meow-intern (symbol-name state) "-mode") 1))
+      (meow--disable-current-state)
+      (meow--switch-state state t))
 
      ;; if key A is bound to a self-insert command
      ((progn
--- a/meow-core.el
+++ b/meow-core.el
@@ -138,18 +138,22 @@ Note: When this function is called, NORMAL state is already enabled.
 NORMAL state is enabled globally when `meow-global-mode' is used.
 because in `fundamental-mode', there's no chance for meow to call
 an init function."
-  (let ((state-to-modes (seq-group-by #'cdr meow-mode-state-list)))
+  (let ((state (meow--mode-get-state)))
     (cond
      ;; if MOTION is specified
-     ((apply #'derived-mode-p (mapcar #'car (alist-get 'motion state-to-modes)))
+     ((eq state 'motion)
       (meow-normal-mode -1)
       (setq meow--current-state nil)
       (meow--save-origin-commands)
       (meow-motion-mode 1))
 
      ;; if NORMAL is specified
-     ((apply #'derived-mode-p (mapcar #'car (alist-get 'normal state-to-modes)))
+     ((eq state 'normal)
       nil)
+
+     (state
+      (meow-normal-mode -1)
+      (funcall (meow-intern (symbol-name state) "-mode") 1))
 
      ;; if key A is bound to a self-insert command
      ((progn
--- a/meow-core.el
+++ b/meow-core.el
@@ -150,19 +150,7 @@ an init function."
 
      (state
       (meow--disable-current-state)
-      (meow--switch-state state t))
-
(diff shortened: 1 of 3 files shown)


meow-helpers.el
```diff
@@ -226,5 +226,18 @@ This function produces several items:
 					    "meow--update-cursor")
 			    ,keymap))))
 
+(defun meow--mode-get-state ()
+  "Get initial state for current major mode."
+  (let ((modes (alist-get nil
```
eshrh, line 231: I think this first let binding is a bit hard to read. What do you think about this:

```
(cl-remove-if-not (lambda (mode)
                    (derived-mode-p (car mode)))
                  meow-mode-state-list)
```

eshrh:
Is this working as intended? As far as i can tell, your `meow--mode-get-state` traces back up the mode derivation until it hits the first non-derived mode that has an entry in `meow-mode-state-list`. doesn't it make more sense to just pick the closest mode (derived or not) that has an entry in `meow-mode-state-alist`?

meow-helpers.el
Xervon: I removed the first binding entirely and just used meow-mode-state-list directly.

Xervon:
> Doesn't it make more sense to just trace up derivations to find the closest mode (itself derived or not) that has an entry in `meow-mode-state-list`?

This is what this does since `throw` should jump out of `while` on the first found mode.

eshrh:
I think that there's still the case where the major mode is not derived and not in the list. here, it should return some default value like `normal` instead of erroring because it returns nil.

Also, what do you think of this rewrite of your get-state function?
```emacs-lisp
(defun meow--mode-get-state (&optional mode)
  "Get initial state for current major mode."
  (let ((mode (if mode mode major-mode))
        (state (alist-get mode meow-mode-state-list)))
    (cond
     (state state)
     ((derived-mode-p mode) (meow--mode-get-state
                             (get mode 'derived-mode-parent)))
     (t 'normal))))
```
Personally, it's a bit easier to follow because theres no nonlocal exit (catch/throw). Also, it makes the function more useful because you can query the closest parent of any mode instead of just the major one.

Xervon:
> I think that there's still the case where the major mode is not derived and not in the list. here, it should return some default value like `normal` instead of erroring because it returns nil.

`meow--enable` already has a fallback to `motion` state, though I do agree that `meow--mode-get-state` might be the better place to implement the fallback instead of doubling code `meow--enable`.

> Personally, it's a bit easier to follow because theres no nonlocal exit (catch/throw).

I agree, but would your rewrite actually work? As far as I can tell `derived-mode-p` checks if the current `major-mode` is derived from one of it's arguments. What do you think of this?

```elisp
(defun meow--mode-get-state (&optional mode)
  "Get initial state for current major mode."
  (let ((mode (if mode mode major-mode))
        (state (alist-get mode meow-mode-state-list))
        (parent (get mode 'derived-mode-parent)))
    (cond
     (state state)
     (parent (meow--mode-get-state parent))
     (t 'normal))))
```

eshrh:
Sorry, I didn't notice. Anyway, that looks good to me!

Xervon:
I set `motion` state as the fallback since that was the fallback before my changes.

MagicRB:
looks like a handy PR, some buffers that i want in motion state start in normal state for some reason. For example the LSP describe thing at point buffer.

Xervon:
> looks like a handy PR, some buffers that i want in motion state start in normal state for some reason. For example the LSP describe thing at point buffer.

You can already do this. This PR just adds the ability to do the same for all the other states. Just add `(MODE . 'motion) to `meow-mode-state-list` and you should be good.

Xervon:
@eshrh 

> I think that there's still the case where the major mode is not derived and not in the list. here, it should return some default value like `normal` instead of erroring because it returns nil.

On further thought, doesn't this prevent the test whether `a` is bound to `self-insert`? This was previously used to determine, if meow should start in `normal` or `motion` state if the mode wasn't in `meow-mode-state-list`.

I feel like `meow--mode-get-state` should either
1. return `nil` if the mode wasn't found or
2. implement the test for `a` directly to inform the fallback state.

meow-core.el
```diff
@@ -138,19 +138,23 @@ Note: When this function is called, NORMAL state is already enabled.
...
       (meow-motion-mode 1))
 
      ;; if NORMAL is specified
-     ((apply #'derived-mode-p (mapcar #'car (alist-get 'normal state-to-modes)))
+     ((eq state 'normal)
       nil)
 
+     (state
+      (meow-normal-mode -1)
+      (funcall (meow-intern (symbol-name state) "-mode") 1))
```
eshrh, line 156: This is brittle. Look up the mode symbol from `meow-state-mode-alist`

eshrh:
Sorry, i didn't have a ton of time to test this out.

Looking at the code now, I think the test for a single letter key is fairly inflexible. Maybe better is to test every letter key, and make sure they are ALL self insert. What do you think about that? I've been affected by the "a" letter check failing one time.

I think that all this stuff should definitely be implemented in meow--mode-get-state though. meow--enable is a little bit unwieldy and hard to read, it would do the file some good to extract logic into a neater function.

eshrh:
I can think of one case where checking all letters will fail. If the user is using lispy, and the first char of the buffer happens to be an open paren, then some keys will trigger the lispy command. Maybe its better to check if ANY of the 26 letters toggles self-insert?

eshrh:
Works perfectly as far I can tell. Thank you!

PR #223, opened 2022-04-23
Fix meow-open-above

krdzo:
When you have `tab-always-indent` set to `'complete` then there is a bug where `meow-open-above` sometimes tries to complete at point, so this is a fix for that.


DogLooksGood:
LGTM, thank you!

DogLooksGood:
LGTM, thank you!

PR #232, opened 2022-05-01
fix customization declarations

ceed0:
I also noticed problems with documentation. For example `meow-mode-state-list` states that 

> The init-state can only be `motion' or `normal'

But actually it can be any state. Or is it intentional?

### DIFF
--- a/meow-var.el
+++ b/meow-var.el
@@ -44,7 +44,7 @@ This will affect how selection is displayed."
   '(markdown-mode org-mode)
   "A list of major modes where after command expand should be disabled."
   :group 'meow
-  :type 'list)
+  :type '(repeat sexp))
 
 (defcustom meow-selection-command-fallback
   '((meow-change . meow-change-char)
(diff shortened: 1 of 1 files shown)


eshrh:
Looks good! As for `meow-mode-state-list`, the documentation is outdated after #217.

ceed0:
Alright then, maybe it would be better if you rewrite it, because my English isn't that great.
Also, I was thinking, should repo documentation be converted to info format?

eshrh:
Sure, I'll take care of it.

I was also thinking it could be neat to have our docs in info, because then it could be hosted both online and inside emacs. I think it might take a lot of work though.

DogLooksGood:
The problem is info doesn't support dynamic content, and the key bindings in Meow is not fixed. So in info we can only use the command name, not that much readable.

ceed0:
>info doesn't support dynamic content

That's not the case, as a random example here is a doc for [imenu](https://github.com/emacs-mirror/emacs/blob/3af9e84ff59811734dcbb5d55e04e1fdb7051e77/doc/emacs/programs.texi#L253) and here how it looks in emacs 
image

PR #238, opened 2022-05-07
Initial proposal for a Vim Comparison

snowiow:
Hi, I started creating a small comparison guide to Vim. It's mostly build up like a cheat sheet for vim users to find out quickly things they know how to do in Vim and how they could be achieved in meow. 
I also added some things which are still empty, because those are the things I used frequently in Vim but couldn't find how they are done in meow yet. 

I would also like to keep this PR open for a bit so that other people can contribute as well.

### DIFF
--- a/VIM_COMPARISON.org
+++ b/VIM_COMPARISON.org
@@ -47,76 +47,82 @@ each command there.
 
 * Movement
 
-| Vim          | Meow                                      | Description                                                                                       |
-|--------------+-------------------------------------------+---------------------------------------------------------------------------------------------------|
-| h            | ~meow-left~                                 | character left                                                                                    |
-| j            | ~meow-next~                                 | next line                                                                                         |
-| k            | ~meow-prev~                                 | previous line                                                                                     |
-| l            | ~meow-right~                                | character right                                                                                   |
-| w            | ~meow-next-word~                            | next word                                                                                         |
-| W            | ~meow-next-symbol~                          | next symbol (as determined by the [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Syntax-Tables.html][syntax table]])                                                   |
-| b            | ~meow-back-word~                            | last word                                                                                         |
-| B            | ~meow-back-symbol~                          | last symbol                                                                                       |
-| e            | ~meow-mark-word~                            | end of word                                                                                       |
-| E            | ~meow-mark-symbol~                          | end of symbol                                                                                     |
-| ge           | =meow-previous-word= - 1 - =meow-next-word=    | end of previous word                                                                              |
+| Vim          | Meow                                          | Description                                                                                       |
+|--------------+-----------------------------------------------+---------------------------------------------------------------------------------------------------|
+| h            | ~meow-left~                                   | character left                                                                                    |
+| j            | ~meow-next~                                   | next line                                                                                         |
+| k            | ~meow-prev~                                   | previous line                                                                                     |
+| l            | ~meow-right~                                  | character right                                                                                   |
+| w            | ~meow-next-word~                              | next word                                                                                         |
+| W            | ~meow-next-symbol~                            | next symbol (as determined by the [[https://www.gnu.org/software/emacs/manual/html_node/elisp/Syntax-Tables.html][syntax table]])                                                   |
+| b            | ~meow-back-word~                              | last word                                                                                         |
+| B            | ~meow-back-symbol~                            | last symbol                                                                                       |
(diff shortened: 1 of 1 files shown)


DogLooksGood:
mark-word and mark-symbol are used to move to the end of word/symbol. 

jump to matched paren can be partly solved by meow-block. 

meow-search will respect the direction, that's how you search backward.


emacs has builtin repeat command. 

emacs has builtin commands for upper/lower/capitcal case.

VIM_COMPARISON.org
```diff
@@ -0,0 +1,99 @@
...
+In comparison meow has the clear advantage when creating the selection, you already get a visual feedback and then can decide which verb to execute. Or you abort if you selected something wrong. In Vim on the other hand you commit on the verb right away, then you need to count blindly the words (which is harder the more things you want to act on) and finally do the action.
+
+However since many people who want to give meow a try, come from a Vim / Evil background, we want to give an overview of typical Vim keybindings and how they can be achieved in meow.
+
+* Movement
```
Linerre, line 14: Would it be possible to turn the following key comparison into an org table? 

At first look, I thought the leftmost column of keys also work in meow. So it could be a bit misleading. 

Something like the below might be better:
| Vim key | Meow Command |
| --------| ------------------|
| `h`        | `meow-left`|

What do you think?

snowiow: I created multiple tables now, one for each headline. It's much more readable now indeed 👍

snowiow:
> mark-word and mark-symbol are used to move to the end of word/symbol.

ah true, thanks for the hint. I added it now to the comparison

> jump to matched paren can be partly solved by meow-block.

cool, that looks even more powerful because you don't need to be on the exact character

> meow-search will respect the direction, that's how you search backward.

yes, that would resemble the `?` in vim, which is added. Or do you mean one can do `reverse-argument` - `meow-search`. But that doesn't work for me
 
> emacs has builtin repeat command.

But it seems not to do the same as in vim. At least it doesn't repeat my last editing commands, but only the last emacs command.
 
> emacs has builtin commands for upper/lower/capitcal case.

added some examples for this as well. Thanks! 🙂

VIM_COMPARISON.org
```diff
@@ -0,0 +1,122 @@
...
+| 0            | ~meow-line~ - ~meow-reverse~                 | beginning of line                                                         |
+| ^            | ~meow-join~                                  | first none whitespace character                                           |
+| $            | ~meow-line~                                  | EOL                                                                       |
+| f - <symbol> | ~meow-find~ - <symbol>                       | forward to next symbol                                                    |
+| F - <symbol> | ~negative argument~ - ~meow-find~ - <symbol> | backward to last symbol                                                   |
+| t - <symbol> | ~meow-till~ - <symbol>                       | till next symbol                                                          |
+| %            | ~meow-block~                                 | jump to matching paren                                                    |
+|              |                                              | not one to one replacement can jump from anywhere inside the block to end |
```
Linerre, line 38: No sure why the Vim column and Meow column are empty here. Something missing perhaps? Could you double check?

snowiow: I did that on purpose because of the long description but I could put it into a single row

Linerre: I see. To me it looks a bit confusing though. 

I might do sth like this:
1. Add a star `*` or some similar symbol as a superscript, indicating this description has extra explanation somewhere further in this doc. 
2.  Maybe right below this very table, put the extra information/annotation for the description.

What do you think? Feel free to disagree. I'm just thinking this might make the table format more consistent

snowiow: changed it now, looked really strange in github and it takes care of wrapping by itself

Linerre: Yes, it looks strange in github. @DogLooksGood and @eshrh may want to have a final quick review a little bit later. 
It looks good to me.  Thank you @snowiow ^_^

eshrh:
Oh oops i thought I was pushing to my own fork...

Anyway, sorry for the big diff, I filled the paragraphs and reformatted the tables. I also added some bits and rewrote some of the big paragraph in the introduction.

DogLooksGood:
I think we should have an entry for this file. @eshrh Do you have any idea about this, heard you want to update the document.

eshrh:
I'm fine with the file, i didn't expect this to be so long. Can we make a directory for documentation though?

DogLooksGood:
We can, the only problem is that we need to list the directory in melpa recipe, otherwise the file will not be included in the build.

DogLooksGood:
Let's merge this firstly.

PR #247, opened 2022-05-15
fix #246 bug in rendering indicator faces

### LINKED ISSUE #246: bug on meow-*state*-indicator render
It seems no matter how I change the faces of six state indicator, they all render with `meow-insert-indicator`
https://github.com/meow-edit/meow/blob/c4b7df0305530da4a51561556794abaf831a228e/meow-util.el#L161-L169

liuyinz:
Rendering indicator with corresponding faces rather than `meow-insert-indicator `always


eshrh:
Sorry again, I didn't check the repo before i pushed my fix commit. let me know if it's fixed for you on master

PR #248, opened 2022-05-16
Add a new beacon example

zhscn:


meow-tutor.el
```diff
@@ -535,6 +535,31 @@
...
+        read_near(base-(4<<10), (64<<10)+(8<<10), 2<<10);
+        truncate(base);
+        read(0, 3<<20);
+        =>
+        write(t, base, (64<<10)+(1<<10), 'c');
+        read_near(t, base-(4<<10), (64<<10)+(8<<10), 2<<10);
+        truncate(t, base);
+        read(t, 0, 3<<20);
+ 1. Move the cursor to the line below marked -->
+ 2. Select the whole region with \\[meow-line]
```
eshrh, line 549: You need to press 3 after a single meow-line to select the whole region

meow-tutor.el
```diff
@@ -535,6 +535,31 @@
...
+ 3. Press \\[meow-grab] to activate secondary selection
+ 4. Press \\[meow-line] to create fake regions at each line
+ 5. Press \\[meow-start-kmacro-or-insert-counter] to start macro recording
+ 6. Press \\[move-beginning-of-line] to move cursor to beginning of current line
+ 7. Press \\[meow-find] and type '(' to find first left parenthesis
+ 8. Press \\[meow-append] and type 't, '
+ 9. Press \\[meow-cancel] to switch to normal state
+ 10. Press \\[meow-end-or-call-kmacro] to end macro recording and apply to all fake cursors
```
eshrh, line 557: Need to press meow-grab again to exit beacon

meow-tutor.el
```diff
@@ -535,6 +535,31 @@
...
+        write(t, base, (64<<10)+(1<<10), 'c');
+        read_near(t, base-(4<<10), (64<<10)+(8<<10), 2<<10);
+        truncate(t, base);
+        read(t, 0, 3<<20);
+ 1. Move the cursor to the line below marked -->
+ 2. Select the whole region with \\[meow-line]
+ 3. Press \\[meow-grab] to activate secondary selection
+ 4. Press \\[meow-line] to create fake regions at each line
+ 5. Press \\[meow-start-kmacro-or-insert-counter] to start macro recording
+ 6. Press \\[move-beginning-of-line] to move cursor to beginning of current line
```
eshrh, line 553: Instead you can just negative find backwards to the ( character.

eshrh (changes requested):
why do you think we need this in addition to the other two? in my opinion, it would be cool if you could show some of the deeper features of beacon, like meow-swap-grab or something as well.

zhscn:
> why do you think we need this in addition to the other two? in my opinion, it would be cool if you could show some of the deeper features of beacon, like meow-swap-grab or something as well.

Yes, there is no deeper feature, the major idea is using kmacro in beacon, which is introduced in Ex A

> Instead you can just negative find backwards to the ( character.

The result is not what we expect, since every line has multiple ( characters, negative finding will append text on after ( characters.

The correct method is that create a new kmacro and apply it to every line, so `meow-kmacro-lines` also works.

Maybe we need to introduce `meow-find-first` or other helper commands to avoid calling kmacro manually.

---

Actually this example is from https://github.com/meow-edit/meow/blob/master/EXPLANATION.org#case-3 ,

zhscn:
Maybe this snippet is not an ideal example, but I think [EXPLANATION.org#case-3](https://github.com/meow-edit/meow/blob/master/EXPLANATION.org#case-3) should be introduced in tutor

Issue #258, opened 2022-05-31
meow--current-state cannot be used to tell meow state, when none of the state is active 


Inc0n:
`meow--current-state` cannot be used to tell meow state, when none of the state is active.

The usage arise from my custom modeline which relies on `meow--current-state` to tell meow state. When none of the state is active, `meow--current-state` will store the last active meow state, which isn't active.

A potential fix for this is to set `meow--current-state` to `nil`, in the way below:
```
(defun meow--define-state-minor-mode (name
                                      init-value
                                      description
                                      keymap
                                      lighter
                                      form)
  "Generate a minor mode definition with name meow-NAME-mode,
DESCRIPTION and LIGHTER."
  `(define-minor-mode ,(meow-intern name "-mode")
     ,description
     :init-value ,init-value
     :lighter ,lighter
     :keymap ,keymap
     (if (not ,(meow-intern name "-mode"))
         (setq-local meow--current-state nil)
       (meow--disable-current-state)
       (setq-local meow--current-state ',(intern name))
       (meow-update-display))
     ,form))
```

DogLooksGood:
I need to review this. There are exit functions called when exiting a state, in `(meow--disable-current-state)`. I'm not sure whether we can set it to nil here. But there are predicate functions for state detection, e.g. `(meow-normal-mode-p)`.

Inc0n:
Understandably so, I also see that meow--current-state is set to nil in `(meow--disable-current-state)`, which could mean a more appropriate solution to be calling it twice.

```
(if (not ,(meow-intern name "-mode"))
    (meow--disable-current-state)
  (meow--disable-current-state)
  (setq-local meow--current-state ',(intern name))
  (meow-update-display)
  ,form)
```

> predicate functions for state detection

As for predicate function for state detection, my usage involes detecting states flexibly across viper, evil and now meow also, using a state variable would be simpler to use (more desireble), and would be prefered personally.

eshrh:
Pretty sure this is closed by your pr #259

PR #259, opened 2022-05-31
Update meow--current-state to nil, when disabled

Inc0n:
This is to address issue: https://github.com/meow-edit/meow/issues/258.
Note: my tab width is set to 8.

### DIFF
--- a/meow-helpers.el
+++ b/meow-helpers.el
@@ -133,7 +133,8 @@ DESCRIPTION and LIGHTER."
      :init-value ,init-value
      :lighter ,lighter
      :keymap ,keymap
-     (when ,(meow-intern name "-mode")
+     (if (not ,(meow-intern name "-mode"))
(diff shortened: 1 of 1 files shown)


DogLooksGood:
I'm a bit busy for these days, I'll back to this once I have time.

eshrh:
I'd prefer you used `meow--disable-state` like you suggested in the issue. It just seems better...

Inc0n:
I have just updated the patch correspondingly

eshrh:
oh oops, that's totally my bad, disable-current-state also calls the same mode function, which calls disable-current-state, and so on leading to a lisp nesting stack error. Setq-local commit doesn't have this issue. Could you revert back to your earlier commit or rebase/remove it?

Inc0n:
Fair enough, done.

PR #265, opened 2022-06-14
(cursor pos hack) fix append behavior at end of line

ml729:
Currently, when using `meow-use-cursor-position-hack`, when the cursor is at the end of a line and there is no selection, pressing `meow-append` will cause the cursor to go to the beginning of the next line.
This fix makes it so that we stay on the same line (basically, we do the same thing as `meow-insert` here).

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -447,7 +447,7 @@ This command supports `meow-selection-command-fallback'."
         (meow--switch-state 'motion))
     (if (not (region-active-p))
         (when (and meow-use-cursor-position-hack
-                   (< (point) (point-max)))
+                   (< (point) (line-end-position)))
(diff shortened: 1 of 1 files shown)


DogLooksGood:
Well, in editor like Kakoune, append will go to next line's beginning. It should be like that. 
And I'm not recommend to use `meow-use-cursor-position-hack`, this is an legacy option, added when I tried to make Meow works like Vim/Kak, but then I realized it's definitely a awful idea. Modal editing doesn't mean to be Vim-like.

ml729:
Ok, that makes sense. 
> Modal editing doesn't mean to be Vim-like.

Thanks for this insight.

PR #268, opened 2022-06-24
Remove needless declare-function

syohex:
`melow--execute-kbd-macro` is defined in this file. This fixes the following warning.

```
meow-util.el:51:8: Warning: function ‘meow--execute-kbd-macro’ defined
    multiple times in this file
```


eshrh:
Thanks. This was my miss, i moved the function to meow-utils in b3d7ef62b340a99e3d7653a7681af1adeed54323 but didn't change declarations.

PR #306, opened 2022-08-18
Specify interactive arguments in the interactive form

fbergroth:
This makes the commands repeatable.


eshrh:
Looks like a strict improvement to me.

PR #318, opened 2022-09-02
Skip string-fence syntax class in meow--{inner,bounds}-of-string

fbergroth:
This fixes bounds/inner for python triple quoted docstrings, as well as regular strings in nix-mode.

Not sure if it breaks anything, @DogLooksGood?


DogLooksGood:
LGTM, thanks!

PR #324, opened 2022-09-08
keep track of direction for no-region cases

eshrh:
Take 2 of fixing the issue where reversing doesn't work with no selection. Previous attempt: #216.

Check the commit message for details. This idea was originally implemented in #272, the code in this pull request extracts the relevant parts.

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -128,13 +128,12 @@ The direction of selection is MARK -> POS."
   "Just exchange point and mark."
   (interactive)
   (setq meow--backward-p (not meow--backward-p))
-  (if (region-active-p)
-      (progn
-        (exchange-point-and-mark)
-        (if (member last-command
-                    '(meow-visit meow-search meow-mark-symbol meow-mark-word))
-            (meow--highlight-regexp-in-buffer (car regexp-search-ring))
-          (meow--maybe-highlight-num-positions)))))
+  (when (region-active-p)
+    (exchange-point-and-mark)
+    (if (member last-command
+                '(meow-visit meow-search meow-mark-symbol meow-mark-word))
+        (meow--highlight-regexp-in-buffer (car regexp-search-ring))
+      (meow--maybe-highlight-num-positions))))
 
 ;;; Buffer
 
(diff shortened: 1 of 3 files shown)


meow-command.el
```diff
@@ -123,16 +125,16 @@ The direction of selection is MARK -> POS."
...
+  "Just exchange point and mark."
   (interactive)
-  (meow--with-selection-fallback
-   (exchange-point-and-mark)
-   (if (member last-command
-               '(meow-visit meow-search meow-mark-symbol meow-mark-word))
-       (meow--highlight-regexp-in-buffer (car regexp-search-ring))
-     (meow--maybe-highlight-num-positions))))
+  (setq meow--backward-p (not meow--backward-p))
+  (if (region-active-p)
```
fbergroth, line 131: Use `when` instead of `if+progn`

fbergroth:
> Where did you mean for this to be done? I added it in meow--cancel-selection

That's probably the right thing to do.

meow-command.el
eshrh: missed this when refactoring

eshrh:
Why shouldn't we have direction with no selection?

I think there should be direction in this case, specifically because it lets you reverse on an empty selection. Treating the position of the point/mark as the only source of truth is also imo likely to cause more subtle issues when changing or implementing new functionality like we saw in the beacon-expand pr.

DogLooksGood:
The original idea is about how to reverse an empty line selection? If that's case, I think what we should do is to avoid produce an empty line selection. There are some concerns:

1. Because nowhere displays such direction information, it's a variable that affects following behavior, but you don't know its value.
2. I think there will be a some edge cases.
3. mark/point is the builtin stuff, whenever there's something builtin, we should build around it.

eshrh:
I'm open to any solution that allows you to reverse the empty line selection. I don't understand how to avoid producing one. Can you clarify that?

DogLooksGood:
How about limit the variable only for line selection, init the value when create line selection, use it with condition `selection-type = line` , ignore it elsewhere.

fbergroth:
Is the limitation necessary? Are there any weird behaviors with the current implementation?

eshrh:
Nothing that i've found so far. I'll try some stuff out soon. The limitation is kind of meaningless because afaik, currently line is the only way to make an empty selection. In the future though, implementing some new feature could run into the same problem again.

PR #326, opened 2022-09-09
add FWYR layout

AKmatiAK:
Polish language optimized layout I recently created. [FWYR](https://github.com/AKmatiAK/fwyr-layout)


eshrh:
@DogLooksGood can we make a directory for keyboard layouts? also, a directory for documentation if possible. We might need to change repo recipes but that shouldn't be a big deal

DogLooksGood:
Sure.

PR #327, opened 2022-09-10
Fix two minore issue with cursor updating.

pRot0ta1p:
* meow--set-cursor-color: call set-cursor-color only when cursor color actually changed, because set-cursor-color will trigger a redisplay.

* meow--update-cursor: only update cursor with current window buffer, this solve the issue that flymake or flycheck cause cursor color change during typing.


DogLooksGood:
Thanks!

PR #350, opened 2022-10-16
Add :meow-state keyword for use-package declarations

Olivia5k:
Add simple keyword `:meow-state` added to [use-package](https://github.com/jwiegley/use-package) declarations.  Used to help populate `meow-mode-state-list`, as follows:

```emacs-lisp
(use-package sly
  :meow-state ((sly-inspector-mode . motion)
               (sly-db-mode . motion)))
```

The above would make the `sly-inspector` and `sly-db` modes start in Meow's `motion` state.

Since I know that it's important to the `meow` project to not have any external dependencies, the `meow-use-package` module will only be loaded if `use-package` is installed. I also added a mention in `CUSTOMIZATIONS.org`, so that it's mentioned somewhere in the documentation.  I hope it was the right place to put things.

### DIFF
--- a/CUSTOMIZATIONS.org
+++ b/CUSTOMIZATIONS.org
@@ -358,3 +358,18 @@ change the variable =meow-cursor-type-insert=.
 
 Association list of symbols to their corresponding keymaps. Used
 to generate =meow-*-define-key= helpers.
+
+* Integration to other packages
+** :meow-state use-package keyword
+
+Simple keyword ~:meow-state~ added to [[https://github.com/jwiegley/use-package][use-package]] declarations.  Used to help
+populate ~meow-mode-state-list~, as follows:
+
+#+begin_src emacs-lisp
+(use-package sly
+  :meow-state ((sly-inspector-mode . motion)
+               (sly-db-mode . motion)))
+#+end_src
+
+The above would make the ~sly-inspector~ and ~sly-db~ modes start in Meow's
+~motion~ state.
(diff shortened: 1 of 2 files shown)


DogLooksGood:
Thanks for the contribution. 
We have `meow-shims.el` for all the 3rd party related code.

Olivia5k:
Oh, that makes sense.  I'll move the things there.

Olivia5k:
@DogLooksGood I've moved the code and removed the other files.  I didn't put the loading of this specific code into the `meow--enable-shims` function, since that doesn't make sense for the context in which `use-package` is used.  `use-package` keywords need to be added at compile time, otherwise packages using them won't see them.

Does that work?

DogLooksGood:
Yeah, it looks good to me.

Olivia5k:
```
Error: Aliases should start with the package’s prefix "meow".
Error: "use-package-handler/:meow-state" doesn’t start with package’s prefix "meow".
Error: ‘use-package-handler/:meow-state’ contains a non-standard separator ‘/’, use hyphens instead (see Elisp Coding Conventions).
```

I don't know how to fix those.  The alias cannot have the `meow` prefix, since it's aliasing something outside of `meow`, and it has to use the `/` as a separator because that's what `use-package` uses.  What do we do?  Can we disable those checks for this code specifically?

DogLooksGood:
Let me do some research here :)

DogLooksGood:
I simply has no idea about how to get rid of those messages. Probably there's a way to let the linter to ignore it?

PR #355, opened 2022-10-26
Fix documentation for `meow-define-state`

yilinwei:
The documentation was incorrect for `meow-define-state`. I decided to keep the existing behaviour and change the documentation rather than change the behaviour.


eshrh:
Wow, can't believe I missed that. It originally was :form, but then i changed the code to be more like `define-minor-mode` Thanks!

PR #369, opened 2022-11-20
tutor correction on line 582

bpsundar:


DogLooksGood:
Hi, this is expected. `SPC a` will translate to `C-c a`. 

Meow has a configurable leader map used as fallback behavior in KEYPAD state. By default, it's `mode-specific-map`. So `SPC a` will be translated to `a` in the `mode-specific-map`, which is `C-c a`.

PR #372, opened 2022-11-21
Add shim for `view-mode`

mmarshall540:
Regarding discussion: #368

Switches Meow-state to `meow-insert-mode` when using `view-mode`.  Switches back after exiting `view-mode`.

I am not sure why the extra 2 empty commits appear.  If there's a way to prevent that, please let me know.


mmarshall540:
With regard to the extra 2 commits, one is a copy of a commit from an older merged PR (and thus, makes no changes). The other is from when I merged the more recent changes from the Meow git repo into my repo.  Thus, only the "Add shim for `view-mode`" commit makes any changes.

mmarshall540:
Per comment on the linked discussion, I no longer think this is a good idea.

It's more effective to use `read-only-mode` and add any desired `view-mode` bindings to the `meow-normal-state-keymap`.

PR #375, opened 2022-12-02
Enable secondary selection via prefix arg in meow-*-thing

fbergroth:
I think this is useful if you want to enable secondary selection, but not move point.

For instance, you can now quickly rename a symbol at point in a function by `W - . d c <new name> ESC`.

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -1286,31 +1286,44 @@ To search backward, use \\[negative-argument]."
    (alist-get cmd meow-thing-selection-directions)
    'forward))
 
-(defun meow-beginning-of-thing (thing)
-  "Select to the beginning of THING."
-  (interactive (list (meow-thing-prompt "Beginning of: ")))
-  (save-window-excursion
-    (let ((back (equal 'backward (meow--thing-get-direction 'beginning)))
-          (bounds (meow--parse-inner-of-thing-char thing)))
-      (when bounds
-        (thread-first
-          (meow--make-selection '(select . transient)
-                                (if back (point) (car bounds))
-                                (if back (car bounds) (point)))
-          (meow--select))))))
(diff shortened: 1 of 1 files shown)


DogLooksGood:
I like this idea! I'll test it a little bit.

DogLooksGood:
I think create grab and preserve the previous cursor position is a good idea. But I want to add this feature to grab. 
I propose this implementation of `meow-grab`, how do you think?

So you can use whatever command to create the selection, then use `-`, `grab` to grab and move cursor to the original position.

```
(defun meow-grab (arg)
  "Create secondary selection or a marker if no region available.

Use prefix argument to preserve the cursor position."
  (interactive "P")
  (if (region-active-p)
      (secondary-selection-from-region)
    (meow--cancel-second-selection))
  (if arg
      (let ((original-pos (caddar (last meow--selection-history))))
        (when original-pos
          (meow--cancel-selection)
          (goto-char original-pos)))
    (meow--cancel-selection)))
```

fbergroth:
I think both yours and mine are good ways to do it.

I almost always want to get back to my previous position after grabbing, so I would probably use `-`, `grab` more often.
In my solution, I like that `- . b` does not jump to the beginning of the buffer. I also like that beacons directly get activated from the previous motion.

I think that the prefix arg to `meow-*-thing` is a natural extension.

Do you think they are exclusive? Can't we add both?

DogLooksGood:
There aren't much DWIM commands in meow. All DWIM commands are made with "selection fallback behavior", by combining two commands. I would say meow is against DWIM-style, so a command creates selection shouldn't also create secondary-selection.

fbergroth:
I have a better idea for this, I'll update it in some time.

fbergroth:
@DogLooksGood see #376.

PR #376, opened 2022-12-03
Add meow-grab-next

fbergroth:
I think this is useful if you want to enable secondary selection, but not move point.

For instance, you can now quickly rename a symbol at point in a function via beacons by doing `mark-symbol`, `grab-next`, `bounds-of -thing defun`.

### DIFF
--- a/meow-beacon.el
+++ b/meow-beacon.el
@@ -37,10 +37,12 @@
 (declare-function meow--selection-fallback "meow-command")
 (declare-function meow--make-selection "meow-command")
 (declare-function meow--select "meow-command")
+(declare-function meow--pop-selection "meow-command")
 (declare-function meow-beacon-mode "meow-core")
 
 (defvar-local meow--beacon-overlays nil)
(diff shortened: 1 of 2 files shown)


fbergroth:
Perhaps bind `meow-grab-next` to `M`?

fbergroth:
> 1. How to cancel this flag?

It should handle the case if region is not active, then you can cancel with `g` or any key that doesn't make a selection.
 
> 2. What's the benefit, compare to use prefix argument. One fewer key?

It's general and works for any command such as `meow-line` or the builtins `mark-defun`. This is not DWIM, neither was the prefix arg. But you seemed to dislike the overloaded behavior.

DogLooksGood:
Yeah, it's not DWIM, but it uses a different order, it's "grab, select", instead of "select, grab".
Another concern is introducing a variable in this case may not be a good idea, cancel a flag is tricky in some cases, for example, `C-g`.

PR #432, opened 2023-04-09
Fix #277: use input method only in insert state

### LINKED ISSUE #277: Single-letter keybindings don't work when an input method is active
Minimal init.el:
```
(add-to-list 'load-path "~/.config/emacs/elpa/meow-20220629.520")
(set-input-method 'russian-computer)
(require 'meow)
(meow-normal-define-key '("j" . meow-next))
(meow-global-mode 1)
```

Steps to reproduce:
1. Run emacs.
2. `C-x b *scratch*`
3. `j` works as expected if the input method is inactive. Once it is activated with `C-\`, pressing `j` raises an error "о in undefined".

Expected behavior: `j` works both when the input method is active and it is not.

akho:
Fix is based on code from modalka, as mentioned in the issue discussion.


DogLooksGood:
We actually have a `meow-shims.el` to deal with all incompatibility issues.  I propose to have it there.

akho:
@DogLooksGood 
I’ve force-pushed a change moving things to meow-shim; thank you for the pointer

Issue #431, opened 2023-04-09
Entering regex into `meow-visit` with `meow-visit-sanitize-completion` enabled does not work


akho:
When `meow-visit-sanitize-completion` is enabled, regexps entered in `meow-visit` minibuffer prompt are sanitized, so you  cannot actually enter a regexp, only search for symbols. This is because `meow--prompt-symbol-and-words` uses `regexp-quote` for its output when `meow-visit-sanitize-completion` is enabled.

I think the intention was just to have a cleaner view of completion candidates, not to prevent the user from entering regexes.

Removing the second check and the quoting seems to work for me, but the behavior was introduced in c29239b with message “bugfix for meow-visit”; I’m reluctant to submit a pull request without knowing what the bug was.

This is what seems to work for me:

```
(defun meow--prompt-symbol-and-words (prompt beg end)
  "Completion with PROMPT for symbols and words from BEG to END."
  (let ((completions))
    (save-mark-and-excursion
      (goto-char beg)
      (while (re-search-forward "\\_<\\(\\sw\\|\\s_\\)+\\_>" end t)
        (let ((result (match-string-no-properties 0)))
          (when (>= (length result) meow-visit-collect-min-length)
            (if meow-visit-sanitize-completion
                (push (cons result (format "\\_<%s\\_>" (regexp-quote result))) completions)
              (push (format "\\_<%s\\_>" (regexp-quote result)) completions))))))
    (setq completions (delete-dups completions))
    (let ((selected (completing-read prompt completions nil nil)))
       (or (cdr (assoc selected completions))
           selected))))
```

Edit: mentioning @shitianshu as the author of the original commit.

DogLooksGood:
That's me, but damned, I guess I forget the reason. I think it's an issue when you trying to enter a raw non-regexp search.  Because the default candidates is sanitized, and people may not realize it's a regexp.

akho:
But the result is that you either see ugly completion options, or lose regexp search (…in meow-friendly bindings), and the docs are incorrect. Maybe it would make sense to set two options, for completion candidates and for raw/regexp?

Though I don’t think anyone actually prefers unsanitized completion candidates, and raw/regexp should really be different commands.

(on a side note, it would be really nice to have incremental search as a motion, with selection and number hints)

DogLooksGood:
[...]

PR #433, opened 2023-04-10
Clarify docstring for meow-visit #431

akho:
Default behavior now is you either select one candidate (word or symbol in buffer), or search for raw string. The regexp quote will prevent it to be treated as regexp. Docstring should not give an impression that `meow-visit` searches for regexps.

The dual function of `meow-visit-sanitize-completion` is also explained in better detail.


DogLooksGood:
Thank you!

PR #436, opened 2023-04-15
Fix Insert Mode example instructions

sltong:
Step 4 instructs the user to press `\\[meow-join] \\[meow-append]` to move to the beginning of the line, when really the command should be `\\[meow-join] \\[meow-insert]`.


DogLooksGood:
I think it should be `meow-append`.

PR #501, opened 2023-09-03
Add a new prompter for `meow-visit`.

okamsn:
These changes make type what is sought in `meow-visit` more like Isearch and Anzu. It is still a work in progress, but it seems to be working.

The change adds a user option to configure a "prompter" which is used to read text from the minibuffer. It also adds a new prompter that highlights matches from point to the top or bottom of the window, depending on the argument to `meow-visit`. If point is inside the secondary selection, then matches within the selection are also highlighted. I find this useful for running `meow-visit` to place the beacons when `meow-visit-sanitize-completions` is nil.

Are you interested in adding something like this? If so, what would you like to see changed?


DogLooksGood:
Hey, I just gave it a try, and it looks cool.  The `buffer-highlight` style seems to be good and I'd like to have this change.

BTW, should the window view scroll to the occurrence which is not in current window view?

okamsn:
I'm now thinking about whether the new other way of doing it could just use Isearch, instead of me re-inventing a worse Isearch with less understanding.

Would something like a user option for a function that returns the beginning and end of the match be acceptable? I see that in https://github.com/meow-edit/meow/discussions/175 there is talk about how to integrate with other search packages. I can check what level of effort would be needed to work with the other packages.

VitalyAnkh:
@okamsn How's the status of this PR now?

DogLooksGood:
There's another PR(#560) to replace the search functionality in meow with ISearch with some tight integration. I modified a lot and put it in a branch [isearch](https://github.com/meow-edit/meow/tree/isearch). Generally I think it's a better approach than implementing a new search. However there are still some issues unsolved, I left them in the commit message.

PR #503, opened 2023-09-06
Add shim for sly

jgarte:
Hi,

This adds a shim for the sly debugger.

For use in the following type of buffer on the left:

image

### DIFF
--- a/meow-shims.el
+++ b/meow-shims.el
@@ -232,6 +232,25 @@ Argument ENABLE non-nil means turn on."
     (remove-hook 'cider--debug-mode-hook 'meow--cider-debug-hook-function)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; sly (db)
+
+(defvar meow--sly-setup nil)
+
+(defun meow--sly-debug-hook-function ()
+  "Switch meow state when entering/leaving sly-db-mode."
+  (if (bound-and-true-p sly-db-mode-hook)
+      (meow--switch-to-motion)
+    (meow--switch-to-motion)))
+
+(defun meow--setup-sly (enable)
+  "Setup sly.
+Argument ENABLE non-nil means turn on."
+  (setq meow--sly-setup enable)
+  (if enable
+      (add-hook 'sly-db-hook 'meow--sly-debug-hook-function)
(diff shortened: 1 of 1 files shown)


jgarte:
I just noticed that I need to do the same thing for `sly-stickers-shortcut-mode`.

How do you suggest I structure that?

DogLooksGood:
I'm not familiar with sly, but I would like to have the PR.

DogLooksGood:
I'm not familiar with sly, but I would like to have the PR.

jgarte:
This package will also need a shim for its mode:

https://github.com/joaotavora/sly-macrostep

I just tested it and meow gets in the way.

jgarte:
> I saw there's only the code or `sly` itself, do you have more commits before we merge it?

What I have here currently works only for the debugger restart window. I think what is currently here is useful but not comprehensive.

I'm not sure at the moment how to structure the code to cover all the other sly modes where we might want to also do the shim code.

Any thoughts?

DogLooksGood:
It's not likely we need shim code for all the modes.  For a major mode derived from `special-mode`, motion state will be used by default.  The same to all text editing major modes. 

The only case we have to handle is when we need a mode switch between normal/motion state due to some special minor mode.

jgarte:
> It's not likely we need shim code for all the modes. For a major mode derived from `special-mode`, motion state will be used by default. The same to all text editing major modes.
> 
> The only case we have to handle is when we need a mode switch between normal/motion state due to some special minor mode.

@DogLooksGood Is there an existing example of what you are describing that I can follow?

DogLooksGood:
I mean, for most major modes, you don't need any shim code.

DogLooksGood:
Do you think this PR is ready for merge?

jgarte:
> Do you think this PR is ready for merge?

Hi,

It works for the sly debugger [restart](http://www.lispworks.com/documentation/HyperSpec/Body/09_adb.htm) view. It does not cover all of sly though.

I think it is ready to merge at least for the sly debugger restart menu.

jgarte:
Hi,

When do you think that this change will land in ELPA?

Just asking because then I can remove it from my Emacsaroni 🍜 

https://git.sr.ht/~whereiseveryone/emacsaroni/tree/master/item/site-lisp/meow-sly-shim.el

DogLooksGood:
it seems that we have to cut a release to make it updated?

jgarte:
Yep, a release party 🎉 ;()

PR #509, opened 2023-09-23
Add `meow-replace-pop`.

okamsn:
This command, when run after `meow-replace` or itself, replaces the just inserted text with the next item in the kill ring, without rotating the kill ring.

- Add command `meow-replace`.
- Add variables `meow--replace-pop-index` and `meow--replace-start-marker`.
- Modify commands `meow-replace`, `meow-replace-save`, and `meow-replace-char` to set the marker before inserting the replacement text.

The marker was needed to support `meow-replace-char`, which doesn't use a region. Otherwise, one could use `(mark t)` to get the mark's position.

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -541,6 +541,7 @@ This command supports `meow-selection-command-fallback'."
      (when (meow--allow-modify-p)
        (when-let ((s (string-trim-right (current-kill 0 t) "\n")))
          (delete-region (region-beginning) (region-end))
+         (set-marker meow--replace-start-marker (point))
          (insert s))))))
 
 (defun meow-replace-char ()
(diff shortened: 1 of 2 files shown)


okamsn:
Why do `meow-replace` and the commands strip a trailing new line from the replacement? I copied that, but I don't understand why it's needed.

DogLooksGood:
Well, the `meow-replace` is designed to replace the selected text with the head of kill-ring. So is here a breaking change?

okamsn:
The command `meow-replace-pop` isn't meant to replace `meow-replace`. It is meant to run run after it. For example, I use `meow-replace-pop` as the fallback for `meow-replace` in `meow-selection-command-fallback`.

Sometimes, I want to replace the region with text that is further into the kill ring. By using `meow-replace-pop` as the fallback of `meow-replace`, I can keep hitting `r` to reach the desired text in the kill ring. It is similar to running `yank-pop` after `yank`, except that `meow-replace-pop` does not cycle the kill ring, so that running `meow-replace` later will work as expected.

If the command `meow-replace-pop` is run after `meow-replace`, which part is the breaking change?

DogLooksGood:
Got it, thanks for the explanation. Are we ready for the merge?

okamsn:
Yes, I'm happy with it. I thought about using a prefix argument to reverse the cycling direction, but I now think it's easier to just use `undo`.

DogLooksGood:
Thanks!

Issue #493, opened 2023-08-08
Meow causes magit to drop stashes instead of discarding changes


larstvei:
Hi!

Thank you for providing an emacs-friendly introduction to modal editing!

When placing the cursor on a change in the magit status buffer, pressing `k` should drop the change. If the cursor is on a stash, it should drop the stash.

When using meow, i press `SPC k` instead. However, it often disregards the cursor location (which is on a change), and rather goes to drop the stash.

Disabling `meow-global-mode` and pressing `k` allows me to drop the change, rather than the stash.

The bug only occurs sporadically, and so I am not able to reliably reproduce it. Do you have any idea of what could cause the error?

This is my magit configuration:
```elisp
  (use-package magit
    :bind (:map custom-bindings-map ("C-c m" . magit-status)))
```

This is my meow configuration:
```elisp
  (use-package meow
     :config
     (setq meow-cheatsheet-layout meow-cheatsheet-layout-qwerty)
     (add-to-list 'meow-mode-state-list '(vterm-mode . insert))
     (add-to-list 'meow-mode-state-list '(comint-mode . insert))
     (meow-motion-overwrite-define-key
      '("j" . meow-next)
      '("k" . meow-prev)
      '("<escape>" . ignore))
     (meow-leader-define-key
      ;; SPC j/k will run the original command in MOTION state.
      '("j" . "H-j")
      '("k" . "H-k")
      ;; Use SPC (0-9) for digit arguments.
      '("1" . meow-digit-argument)
      '("2" . meow-digit-argument)
      '("3" . meow-digit-argument)
      '("4" . meow-digit-argument)
      '("5" . meow-digit-argument)
      '("6" . meow-digit-argument)
      '("7" . meow-digit-argument)
      '("8" . meow-digit-argument)
      '("9" . meow-digit-argument)
      '("0" . meow-digit-argument)
      '("/" . meow-keypad-describe-key)
      '("?" . meow-cheatsheet))
     (meow-normal-define-key
      '("0" . meow-expand-0)
      '("9" . meow-expand-9)
      '("8" . meow-expand-8)
      '("7" . meow-expand-7)
      '("6" . meow-expand-6)
      '("5" . meow-expand-5)
      '("4" . meow-expand-4)
      '("3" . meow-expand-3)
      '("2" . meow-expand-2)
      '("1" . meow-expand-1)
      '("-" . negative-argument)
      '(";" . meow-reverse)
      '("," . meow-inner-of-thing)
[...]

PR #517, opened 2023-10-14
Consider local keybindings when moving commands for the Motion state.

okamsn:
- Add the library Compat (https://elpa.gnu.org/packages/compat.html), which makes available the function `keymap-local-lookup` to Emacs older than 29.1. The new `keymap-` functions are recommended over older functions like `lookup-key`.

- Instead of directly binding `H-j` to what was bound to `j`, for example, bind `H-j` to an interactive lambda which first checks whether there is a local binding for `j`.  If there is, use that.  If there is not, then call the orignal command bound to `j`.

This seems to fix using the overrides in Magit buffers (see #493), where the command bound to `H-k` seems to be based on where point is located when `meow--save-origin-commands` is run.

### DIFF
--- a/meow-util.el
+++ b/meow-util.el
@@ -452,6 +452,17 @@ Looks up the state in meow-replace-state-name-list"
    (t nil)))
 
 (defun meow--save-origin-commands ()
+  "Save the commands overridden by the Motion map to modified bindings.
+
+The new key binding, modified by the prefix in
+`meow-motion-remap-prefix', is bound to a command that calls the
+command locally bound to the original key binding, or, if that is
+nil, the original command.
+
+For example, under the default and suggested settings, in a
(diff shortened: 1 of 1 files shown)


DogLooksGood:
Hey, thanks for the contribution.  

We want to keep meow free of dependencies, so we don't want to introduce `compact` library. I guess we can port it if we really need it.

okamsn:
OK, I've removed the use of Compat and copied the relevant portion of the definition.

This change works for me on Emacs 29.1. What else would you like changed?

DogLooksGood:
I'm quite busy these days.  Later I'll check the updates and merge.

norris-young:
This works for me. Will you merge it?

DogLooksGood:
I'll do some test and merge it.

PR #589, opened 2024-05-04
meow--select: don't spam mark-ring

45mg:
For example, without this change, calling `meow-next-word` repeatedly on a selection created by `meow-mark-word` will repeatedly push the position at the beginning of the selection onto the mark-ring. This can quickly overflow the mark-ring with identical positions, and it makes `pop-to-mark-command` a lot more annoying to use.

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -64,7 +64,8 @@ The direction of selection is MARK -> POS."
     (meow--cancel-selection))
   (let ((sel-type (car selection))
         (mark (cadr selection))
-        (pos (caddr selection)))
+        (pos (caddr selection))
+        to-push)
     (if meow--selection
         (unless (equal meow--selection (car meow--selection-history))
           (push meow--selection meow--selection-history))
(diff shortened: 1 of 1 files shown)


DogLooksGood:
I think you inspired me that we should only keep one mark for each type of selection no matter how it gets extended.  So I probably want to have this patch for all commands, and do it in a better way.

45mg:
Glad to hear it! The fact that all the Meow motions create selections means that the mark ring automatically becomes a tool for jumping back to previous edit positions.

How would you feel about an `unpop-to-mark-command` in Meow, to move from older to newer positions in the mark ring? I've already implemented one. With that, combined with this fix, Meow would achieve the functionality of `better-jumper` via built-in Emacs features.

DogLooksGood:
Sorry for the late response, yeah, I would like to learn more about your idea. I think current mark-ring usage wasn't good. I'd like to have an improvement here.

45mg:
> Sorry for the late response, yeah, I would like to learn more about your idea. I think current mark-ring usage wasn't good. I'd like to have an improvement here.

My idea is pretty much what I said in my first comment - since the user moves around and edits using Meow motions, and those motions create selections, the user's movement and edit positions are saved in the mark ring. This means that one can move back to previous positions using `pop-to-mark-command`. I wrote an `unpop-to-mark-command` which then allows you to go to more recent positions. This pair of commands basically models the same functionality as `better-jumper`, or Vim's `Ctrl-O` and `Ctrl-I`.

DogLooksGood:
I think we can achieve this simply by comparing the current selection type and new selection type.

I'm thinking whether we should keep the same type of first choice or last choice.

DogLooksGood:
Fixed in the latest command, also [see](https://github.com/meow-edit/meow/blob/master/COMMANDS.org#meow-pop-to-mark)

Now we have something similar to C-i/C-o in Vim

PR #596, opened 2024-05-18
Allow custom 'word', 'symbol' definitions

45mg:
Allows users to customize the behavior of the commands related to marking and moving by words and symbols.

Make all commands related to words and symbols rely on `forward-thing` and `bounds-of-thing-at-point`. Then, define the 'thing's that these functions are called on in `meow--word-thing` and `meow--symbol-thing`. Now the user can define their own things and make Meow use them by setting these variables.

Meow users may find Emacs' default notions of 'word' and 'symbol' to be unsatisfactory for modal editing. For example, word movements tend to skip over non-alphanumeric characters, meaning that `meow-next-word` and `meow-back-word` will skip them when not expanding a selection. This commit allows users to redefine these notions via `thingatpt`.

With this change, every Meow motion is now customizable. The only exception is `meow-line`; it should not be necessary to change its behavior since Meow already provides `meow-visual-line`.

### DIFF
--- a/CUSTOMIZATIONS.org
+++ b/CUSTOMIZATIONS.org
@@ -358,3 +358,55 @@ change the variable =meow-cursor-type-insert=.
 
 Association list of symbols to their corresponding keymaps. Used
 to generate =meow-*-define-key= helpers.
+
+** meow-word-thing, meow-symbol-thing
+
+The things used by meow for marking/movement by words and symbols, respectively.
+
+The values are 'things' as understood by ~thingatpt~ - symbols that will be
+passed to ~forward-thing~ and ~bounds-of-thing-at-point~, which see.
+
+This means that they must, at minimum, have a function as the value of their
+=forward-op= symbol property (or the function should be defined as
+~forward-SYMBOLNAME~). This function should accept a single argument, a number
+=n=, and should move over the next =n= things, in either the forward or backward
+direction depending on the sign of =n=. Examples of such functions include
+~forward-word~, ~forward-symbol~ and ~forward-sexp~, which ~thingatpt~ uses for
+the =word=, =symbol= and =sexp= things, respectively.
+
+*** Custom =word=, =symbol= definitions
+
+By customizing these variables, you can make Meow use your own definitions for
+=word= and =symbol=. For example, here is how you can get =word= behavior closer
+to Vim's -
+
+#+begin_src emacs-lisp
+(defun forward-vimlike-word (&optional arg)
+  "Alternate `forward-word'. Essentially the same idea as Vim's 'e'."
+  (interactive "^p")
+  (setq arg (or arg 1))
+  (cl-destructuring-bind (sign move-func char-func)
+      (if (>= arg 0)
+          '(1 skip-syntax-forward char-after)
+        '(-1 skip-syntax-backward char-before))
+    (with-syntax-table (standard-syntax-table)
+      (let ((distance sign))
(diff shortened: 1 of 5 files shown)


45mg:
Here is an example of how this could be used to get 'word' behavior closer to Vim's:
```elisp

(defun forward-vimlike-word (&optional arg)
  "Alternate `forward-word'. Essentially the same idea as Vim's 'e'."
  (interactive "^p")
  (setq arg (or arg 1))
  (cl-destructuring-bind (sign move-func char-func)
      (if (>= arg 0)
          '(1 skip-syntax-forward char-after)
        '(-1 skip-syntax-backward char-before))
    (with-syntax-table (standard-syntax-table)
      (let ((distance sign))
        (while (and distance (> (abs distance) 0) (> (* arg sign) 0))
          (setq distance
                (when-let ((next-char (funcall char-func))
                           (next-syntax (char-syntax next-char)))
                  (cond ((eq next-syntax ?w)
                         (funcall move-func "w"))
                        ((eq next-syntax ?\ )
                         (prog1
                             (funcall move-func " ")
                           (forward-vimlike-word sign)))
                        (t
                         (funcall move-func "^w ")))))
          (setq arg (- arg sign)))
        (and distance (> (abs distance) 0))))))

(put 'vimlike-word 'forward-op #'forward-vimlike-word)

(setq meow--word-thing 'vimlike-word)
```

eshrh:
Looks good and works well. This is a useful abstraction to have!

Your old docstring seems better than the one you force pushed? If you feel like it's redundant I think it should at least have some words explaining that you can define your own things to customize behavior. Ideally, you should add a section in CUSTOMIZATIONS.org as well.

I believe you also need to modify the forward-word calls in beacon mode. This section: https://github.com/meow-edit/meow/blob/99e08c92bb5d8a695062ce53e2cffeffd3a058a6/meow-beacon.el#L219-L225

I briefly tested replacing the forward word calls with forward-thing and it seems to work.

45mg:
> Your old docstring seems better than the one you force pushed? If you feel ...

I wrote a new one. I also added a section to `CUSTOMIZATIONS.org`, with my example. It's definitely possible to come up with better word redefinitions - this one generates too many expansion hints because it considers single chars like hyphens between words to also be words - but for now I think it's fine as a proof-of-concept.

> I believe you also need to modify the forward-word calls in beacon mode.

Good catch. Done.

45mg:
I force-pushed to rename the variables to `meow-word-thing` and `meow-symbol-thing`, since they're not internal variables.

45mg:
While fixing a few places where I forgot to replace `forward-word` and `forward-thing`, I realized that a lot of Meow's logic for word and symbol marking/movement is duplicated across multiple functions. So I added a commit that combines all this code into one function that uses `thingatpt`.

45mg:
Force-pushed again to fix some typos (sigh). It should all work now.

45mg:
Force-pushed a small fix - the function to add word beacons no longer requires `meow-forward-thing` to return `t` when it moved successfully.

eshrh:
@DogLooksGood review? looks good to me.

45mg:
Hope this can be merged soon. I had other ideas for PRs but I've been holding off on them, because I'd probably want to use the abstractions introduced in this one...

DogLooksGood:
There could be something wrong, I haven't looked deeply.

The word/symbol forward and backward shouldn't select only the word/symbol.

```
foo| bar baz -> foo| bar| baz

foo| (bar baz) -> foo (|bar| baz)
```

PR #602, opened 2024-06-01
Fixed freezing emacs when no input is provided to the meow-visit minibuffer

bunnyanon:
When `(meow-visit)` is invoked and no input is provided in the minibuffer, Emacs freezes and hogs a CPU core.

this change adds a verification if `text` is empty, and if it is, `visit-point` is set to `nil`, triggering a `visit failed` message for the user.

It also adds a custom error message for this case, it being `Visit <empty> failed`.

### DIFF
--- a/meow-command.el
+++ b/meow-command.el
@@ -1406,14 +1406,12 @@ To search backward, use \\[negative-argument]."
                (end (if (> pos visit-point) (marker-position marker-beg) (marker-position marker-end))))
           (thread-first
             (meow--make-selection '(select . visit) beg end)
-	    (meow--select))
+            (meow--select))
           (meow--push-search text)
-	  (meow--ensure-visible)
-	  (meow--highlight-regexp-in-buffer text)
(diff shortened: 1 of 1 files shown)


eshrh:
Thanks! This is definitely a bug.
 
Would you mind changing your indentation to use spaces instead of tabs? Your patch also currently touches more lines than necessary.

bunnyanon:
Sure! indentations fixed!

eshrh:
Originally I thought that this was probably the best way since it seems that `completing-read` has no built in way to disallow empty input no matter what you do; and that it's impossible to message the user about the error without using sit-for or timers. 

Then I realized just now that you should probably loop until the user gives non-empty input and change the prompt itself which seems like a more user-friendly solution. Failing and stopping like in your patch is usually correct, but in this case there is literally no case where the user wants to visit empty strings -- it's invalid input and therefore shouldn't be allowed at the input stage.

PR #609, opened 2024-06-22
Allow custom `delete-region`, `insert` functions

45mg:
Make it possible to specify which functions are used to delete the region and insert text.

This is needed to get Meow to work with modes like Vterm, that define their own `delete-region` and `insert` functions, without massive code duplication.

See [meow-vterm](https://github.com/45mg/meow-vterm) for an example of this - it requires this patch to work.

### DIFF
--- a/meow-beacon.el
+++ b/meow-beacon.el
@@ -504,7 +504,7 @@ The recorded kmacro will be applied to all cursors immediately."
                   (delete-overlay ov)))))))
 
 (defun meow--beacon-delete-region ()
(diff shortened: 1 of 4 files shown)


45mg:
Force-push to define `meow--insert` and `meow--delete-region`, which should be less annoying than having to `(funcall meow--X-function)` every time.

45mg:
Force-push because the previous one confused `funcall` and `apply`. Oops.

PR #621, opened 2024-08-07
Fix byte-compilation warnings

ZharMeny:
This PR fixes various compilation warnings.

Note: this PR also changes the calling convention for `meow-describe-key`, you may want to handle that in a different way.


eshrh:
Thanks!

It doesn't seem like your change to meow-describe-key affects any code path. Besides, on latest, `describe-key` does not accept the `up-key` arg, it might've been for an earlier version.

PR #637, opened 2024-09-18
Add a shim for Macrostep.

okamsn:
Macrostep shows macro expansions using overlays either in a separate buffer or at the location where the macro is used.  In the overlay, it binds a local map to further expand or collapse macros.

This change takes inspiration for the Beacon state and checks whether point is inside one of the overlays in `post-command-hook`.  However, at least one of the commands to collapse the overlay does not seem to trigger `post-command-hook`, so switching back to the previous state is also done during `macrostep-mode-hook`.

`macrostep` was mentioned in #503 by @jgarte. I am pinging them in case they have thoughts on this shim.

### DIFF
--- a/meow-shims.el
+++ b/meow-shims.el
@@ -251,6 +251,79 @@ Argument ENABLE non-nil means turn on."
     (remove-hook 'sly-db-hook 'meow--sly-debug-hook-function)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; macrostep
+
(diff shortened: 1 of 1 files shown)


eshrh:
looks good. do you want to wait for @jgarte or are you ready for merge?

jgarte:
If you've tested it and it works well then LGTM :)

okamsn:
I had overlooked `keypad` state when switching to `motion` state. That part is fixed. However, the shim won't work with any other states that work like Keypad state, because it switches back to Motion state while in the overlay before the Keypad key sequence is completed. I'm not sure how to handle that case.

Other than that, I think it's good now.

jgarte:
@okamsn Thanks for your contribution. I'll appreciate this feature add.

PR #644, opened 2024-10-11
Add `meow(-backward)-kill-{word,symbol}`

45mg:
Versions of `kill-word` and `backward-kill-word` that respect `meow-word-thing`
and `meow-symbol-thing`. May be bound to M-d and M-DEL in place of `kill-word`
and `backward-kill-word`.


DogLooksGood:
Thanks for the contributing!

PR #650, opened 2024-10-20
Add beacon fake cursors for symbols

Haxxflaxx:
Adds beacon fake cursors for symbols the same way that is done for word.
It still uses `meow--beacon-region-words-to-match ` the same as for word which seems to work fine.


DogLooksGood:
Ops, merged with some issue, but I have it fixed.

PR #656, opened 2024-10-29
Add a shim for grep-edit

gs-101:
`grep-edit` is a major mode introduced in Emacs 31 (introduced in commit [db1eb8a2](https://git.savannah.gnu.org/cgit/emacs.git/commit/etc/NEWS?id=db1eb8a282c1832fd34be049e80dcb1a3b59ade2)) which enables the editing of Grep results, by typing 'e' in '*grep*' buffers. 'C-c C-c' exits this mode and saves changes. It can be used as a built-in replacement of [wgrep](https://github.com/mhayashi1120/Emacs-wgrep).

I can hold this PR back if it causes any issues in earlier versions.


DogLooksGood:
Working on some improvements to simplify the shims at #664.

gs-101:
Thanks, I'll checkout #664 and see if this PR is still necessary.

DogLooksGood:
I think for now it's still necessary. But I want to remove the need somehow.

gs-101:
Second force-push was because the commit had unintentional indentation changes.

PR #662, opened 2024-11-06
Add shim for corfu

Haxxflaxx:
Shim for exiting corfu when exiting insert mode. The same as how company is handled.


DogLooksGood:
Thanks!

DogLooksGood:
We may have a chance to get rid of the shims for company and corfu. check #664

PR #664, opened 2024-11-08
Lower the priority of normal keymap and remove some shims

DogLooksGood:
All keymap in meow had the high priority by setting `emulation-mode-map-alists`.  In normal state, the result is every minor mode single key commands are covered by normal mode commands.  We added a lot shims to work out of this, however it doesn't seems to be necessary to me. 

By lowering the keymap's priority, we can get rid of some shims, e.g. edebug, company, corfu, magit-blame, etc. They just work now.

### DIFF
--- a/meow-core.el
+++ b/meow-core.el
@@ -186,8 +186,6 @@ there's no chance for meow to call an init function."
   (add-to-ordered-list 'emulation-mode-map-alists
                        `((meow-motion-mode . ,meow-motion-state-keymap)))
   (add-to-ordered-list 'emulation-mode-map-alists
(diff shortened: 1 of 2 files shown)


gs-101:
Just tested this PR with `package-vc-install` (and recompiling and restarting just to be safe) and my motion keys just executed `self-insert-command` instead of navigation commands with `grep-edit`. So #656 is still valid (just how the `wgrep` shim still is I'd guess, since you didn't remove it in this commit).

DogLooksGood:
I want to have a automatic normal/motion switching by detecting if there's an self-insert-command bound to single keys. there may be some tricky edge cases, I'll give it a try.

PR #666, opened 2024-11-09
Add eat-eshell shim

Haxxflaxx:
eat-eshell is a mode for eat integration for eshell. This lets you run visual commands directly in eshell instead of having it open it in a separate term buffer.

After a visual command has completed and it returns to eshell it doesn't update the cursor so we should do that on the exit-hook.

As it goes from eshell which is effectively running in line-mode, to eat in semi-char-mode, the experience can be a bit jarring. In semi-char-mode you can't really move around. Moving up and down would instead cycle through the command history. By toggling emacs-mode on entering normal mode makes this experience nicer, though as the some work would be needed to make pasting from normal mode working.

I'm not thrilled about relying on `eat--eshell-semi-char-mode`, maybe this is fragile? This was the only way I could come up with to track which mode we are in.

The rest of eat could probably use a similar shim, but I don't really use the rest so it might be better for someone who do to implement it.

### DIFF
--- a/meow-shims.el
+++ b/meow-shims.el
@@ -469,6 +469,35 @@ Argument ENABLE non-nil means turn on."
       ;; These vars allow us the select through the polymode chunk
       (add-to-list 'polymode-move-these-vars-from-old-buffer v))))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; eat-eshell
+
+(defvar meow--eat-eshell-setup nil)
+(defvar meow--eat-eshell-entered-emacs-mode nil)
+
+(declare-function eat-eshell-emacs-mode "eat")
+(declare-function eat-eshell-semi-char-mode "eat")
+
+(defun meow--eat-eshell-maybe-enter-emacs-mode ()
+  (when (bound-and-true-p eat--eshell-semi-char-mode)
+    (setq meow--eat-eshell-entered-emacs-mode t)
+    (eat-eshell-emacs-mode)))
+
+(defun meow--eat-eshell-maybe-enter-semi-char-mode ()
+  (when meow--eat-eshell-entered-emacs-mode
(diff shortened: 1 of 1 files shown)


DogLooksGood:
Hey, thanks for the PR. Why don't we use char mode instead of semi-char mode. The char mode supports all terminal keys which seems to be better as we already have normal state with emacs mode.

Haxxflaxx:
Good point about using char mode instead. The reason is mainly that it defaults to semi char mode.
If we make it go straight to char mode, then switch that on normal/insert-mode, I think that could work great.

That advice you have there, is that to get into eat-emacs-mode?

To solve the pasting from normal mode; can I wrap meow-yank in semi-char-mode then override it in the keymap only for eat-eshell in any convenient way?

DogLooksGood:
Yeah, I have this advice to get to `eat-emacs-mode` + normal state. Otherwise it will be in semi char mode. 

Add a hook to `meow-normal-mode-hook` would be too expensive as this hook runs every time you switch to normal state. I'd rather to have some hook only in eat buffer. I'm not sure if we have more patches like "C-y" needed.  Since these term/repl modes may have different behaviors on default keybindings(like "C-y"), the way how meow uses underlying command through keybinding may not work well.

Haxxflaxx:
I don't think it's a great idea to override keys in `eat-char-mode` as the purpose of it is pretty much to just send the keys as is. I think it would be better instead to temporarily drop into semi char mode do whatever we need to do then go back to emacs mode

Is there something we can do here with your work with lowering the priority of the keymap? This should allow us to have a shim minor mode that grabs certain keys from normal mode.

Haxxflaxx:
I've updated with changes so that i runs in char mode. I also changed it so that it doesn't rely on some internal variable for keeping track of state.

It is a bit nicer to use as I no longer need to go into char-mode for handling things like f-keys in htop and the like (which is mostly what I use it for. The only thing I don't really know how to handle is how we should send escape. I don't really do that very often outside of running vim in emacs to show that you can, but maybe it would be good to have the alternative.

You could always press "C-[ C-[" to send escape.

PR #668, opened 2024-11-11
Add eat-eshell shim

Haxxflaxx:


Haxxflaxx:
Sorry for opening a new branch. I messed up the branching.

Haxxflaxx:
After using this for a week or so I think it mostly works well. I think one improvement would be to not hard code the transition from semi-char-mode into emacs- or char-mode. I think the better option would be to base that choice would be to base it on the active meow state.

So, if we enter eat from insert mode, then we should end up in char-mode. If we do it in normal mode, then we should use emacs-mode.

I'll throw something together to test tomorrow.

DogLooksGood:
Ohh, I added some comments to the previous PR, they lost. 

Well, when adding hook to `meow-insert-enter/exit-hook` we should add it buffer locally, by providing the 4th argument to `add-hook`. 

Also I think we should not do this advice, as it is global runs whenever the command is called, but for eat only.

We should avoid global modification whenever possible.

Haxxflaxx:
Yes, this makes sense. I wasn't aware about adding buffer local hooks. This would allow me to remove the maybe switch mode functions.

I guess instead of having the advice I could trigger the mode override on `eat-eshell-exec-hook`, and that way it would only run once on enter. This would allow a user to easily use semichar mode if needed.

I'll try implementing  this and we'll see where it gets us.

Haxxflaxx:
I've made some improvements and the result is much nicer. 

I couldn't come up with a way to avoid adding advice to `meow-yank` to handle the mode switching required for pasting from `eat-eshell-emacs-mode`.

DogLooksGood:
I think we should just remove the advice.

Haxxflaxx:
Could this be handled instead by making the shim a minor-mode with a keymap overriding the `meow-normal-state-keymap`? Then users could bind whatever they want into it.

The shim could provide a function for doing the emacs-mode pasting that they could bind to it if they want it.

Haxxflaxx:
I realized that if I stop overthinking it and just bind my fancy mode switch yank function in the `eat-eshell-emacs-mode-map` then this just works.

This should be a much simpler and nicer version of the shim.

DogLooksGood:
I don't think we should make it complicated and provide any options/docs from shims. If use you want to do copy/paste with kill-ring/clipboard they can bind another command just like how people do in normal terminal emulator with `C-S-c` and `C-S-v`.  Just keep it simple.

DogLooksGood:
I think to paste, there's no need to switch to semi char mode first, `eat-yank` works just fine.

Haxxflaxx:
It won't work in emacs-mode as the buffer is read-only. At least this doesn't work for me. The mode swapping in the shim paste function is to work around this.

DogLooksGood:
In emacs-mode the cursor can go anywhere, I don't think it's good to call eat-yank there. If we want to adapt it, we should add advice to `eat-yank`.

Haxxflaxx:
When moving the cursor in emacs mode and pasting with `eat-yank` the content will go wherever the underlying terminal thinks the cursor is. This is the same behavior as moving the cursor then entering insert at typing.

When doing line-editing this works kind of as expected.

It is a bit strange in visual programs like say nano, where it will still paste at the original cursor but there won't be any indicator of where that is.

I think about this more like copy mode in tmux, where it isn't really for editing, compared to running shell-mode where you can do whatever you want in the buffer and only the last line is handled in a special way.

DogLooksGood:
I don't want to provide any commands from this shim module as I never consider shimming everything is a good way to go, and a stable solution.  So I prefer to let users handle it for now.

Haxxflaxx:
Seems reasonable.

Haxxflaxx:
Moved out the function. Now the shim only handles the mode override and restoring the correct cursor shape after exit.

DogLooksGood:
Thank you very much!

Haxxflaxx:
If line-mode is implemented at some point for eat-eshell, [feature request](https://codeberg.org/akib/emacs-eat/issues/121), then this would probably be a nicer thing to use for normal mode as it would allow you to edit the prompt line as per normal meow usage.

If we add a shim for eat itself I would say this is a better mode to use than emacs-mode.

PR #683, opened 2024-12-01
meow-thing-register support keyword pair-regexp &  add new thing code-block

jixiuf:
meow-thing-register support keyword `pair-regexp`

`pair-regexp` works similarly to `pair`, but while `pair` uses ordinary strings,
`pair-regexp` utilizes regular expressions.

an example 
```
    (meow-thing-register 'code-block
                         '(pair-regexp ("{[\n\t ]*")  ("[\n\t ]*}") )
                         '(pair ("{") ("}")))
```

inner  thing of code-block 
look like:
图片
instead of 
图片

this is usable when I want to call `meow-grab` on all fields of a struct.

### DIFF
--- a/meow-thing.el
+++ b/meow-thing.el
@@ -204,6 +204,13 @@ Both inner-fn and bounds-fn returns a cons of (start . end) for that thing.")
                       (string-join (mapcar #'regexp-quote tokens) "\\|"))))
     (lambda () (meow--thing-pair-function push-token pop-token near))))
 
+(defun meow--thing-make-pair-regexp-function (x near)
+  (let* ((push-token (let ((tokens (cadr x)))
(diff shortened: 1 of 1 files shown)


DogLooksGood:
`pair-regexp` looks good to me. 

But `code-block` doesn't work in many languages. I remember in kak, it uses a concept called indentation? So basically select the statements starts with same indentation, it's not limited to C-like languages.

jixiuf:
ok you can

> `pair-regexp` looks good to me.
> 
> But `code-block` doesn't work in many languages. I remember in kak, it uses a concept called indentation? So basically select the statements starts with same indentation, it's not limited to C-like languages.

ok I remove the `code-block`

PR #688, opened 2024-12-04
Fix meow--keypad-try-execute

fuzy112:
If you start Emacs with the following configuration and type `SPC p`, an error will occur.

```elisp
(straight-use-package 'cape)
(keymap-global-set "C-c p" 'cape-prefix-map)
```

This is because `cape-prefix-map` is an autoloaded keymap.
`(commandp cmd)` returns t when `cmd` is an autoloaded keymap, but
`(call-interactively cmd)` will raise an error.


DogLooksGood:
Thanks!

PR #700, opened 2024-12-17
Add handling for meow-keypad-self-insert-undefined

Haxxflaxx:
This doesn't fully solve the problem as it doesn't get applied to beacons as expected.
May be due to setting `last-command-event` with a let so it doesn't persist?

It does make it possible to self-insert using the keypad like before.
As it sets `last-command-event` calling prefix arg from keypad also works as expected.

It is now also possible to set `meow-keypad-self-insert-undefined` to nil and bind `self-insert-command` in keypad.

### DIFF
--- a/meow-keypad.el
+++ b/meow-keypad.el
@@ -414,7 +414,8 @@ try replacing the last modifier and try again."
               meow--use-meta
               meow--use-both)
     (let* ((key-str (meow--keypad-format-keys nil))
-           (cmd (meow--keypad-lookup-key (kbd key-str))))
+           (cmd (meow--keypad-lookup-key (kbd key-str)))
+           (last-command-event (string-to-char key-str)))
       (cond
        ((keymapp cmd)
         (when meow-keypad-message (meow--keypad-show-message))
@@ -439,7 +440,9 @@ try replacing the last modifier and try again."
         (meow--keypad-try-execute))
        (t
         (setq meow--prefix-arg nil)
-        (message "%s is undefined" (meow--keypad-format-keys nil))
+        (if meow-keypad-self-insert-undefined
+            (call-interactively 'self-insert-command)
+          (message "%s is undefined" (meow--keypad-format-keys nil)))
         (meow--keypad-quit)
         t)))))
 
--- a/meow-keypad.el
+++ b/meow-keypad.el
@@ -65,10 +65,8 @@
   "Lookup the command which is bound at KEYS."
   (let* ((keybind (if meow--keypad-base-keymap
 		      (lookup-key meow--keypad-base-keymap keys)
-		    (key-binding keys))))
-    (unless (and (meow--is-self-insertp keybind)
-		 (not meow-keypad-self-insert-undefined))
-	  keybind)))
+		      (key-binding keys))))
+    keybind))
 
 (defun meow--keypad-has-sub-meta-keymap-p ()
   "Check if there's a keymap belongs to Meta prefix.
@@ -414,8 +412,8 @@ try replacing the last modifier and try again."
               meow--use-meta
               meow--use-both)
     (let* ((key-str (meow--keypad-format-keys nil))
-           (cmd (meow--keypad-lookup-key (kbd key-str)))
-           (last-command-event (string-to-char key-str)))
+           (last-command-event (string-to-char key-str))
+           (cmd (meow--keypad-lookup-key (kbd key-str))))
       (cond
        ((keymapp cmd)
         (when meow-keypad-message (meow--keypad-show-message))
--- a/meow-keypad.el
+++ b/meow-keypad.el
@@ -390,6 +390,7 @@ Returning DEF will result in a generated title."
   "Execute the COMMAND.
 
 If there are beacons, execute it at every beacon."
+  (setq last-command-event last-input-event)
   (if (meow--keypad-in-beacon-p)
       (cond
        ((member command '(kmacro-start-macro kmacro-start-macro-or-insert-counter))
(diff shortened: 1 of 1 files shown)


Haxxflaxx:
I also noticed there is something weird going on with undo when doing keypad self insert.

Haxxflaxx:
I've fixed the issue where the command wasn't run on beacons, and also added an undo boundary before running it.

Haxxflaxx:
At this point I believe the keypad works the same for self insertions as it did before the refactor, with the exception that you can now explicitly set self insertions in the keypad map instead of having the variable toggle it for all unbound keys.

DogLooksGood:
I think we have to be caution about this change, it may break things.

Haxxflaxx:
> I think we have to be caution about this change, it may break things.

Sounds reasonable. Is there anything in particular that you are worried about?

DogLooksGood:
I'm thinking maybe we can achieve the goal with another method. If users want to call self-insert-command in a specific use case.

Haxxflaxx:
Can you elaborate on what you mean? Do you mean instead of setting `last-command-event`, we update some other part of the keypad which result in the correct value being set so we don't have to?

Haxxflaxx:
Ok, so after thinking a bit more about this I am 1. amazed that this works at all, and 2. fairly certain that I should have done something like this instead:
```emacs-lisp
(setq last-command-event (aref (kbd (meow--keypad-format-keys)) 0))
```

This also makes me a bit uncertain `meow--keypad-format-keys` is the best choice as I would need to know if I should add a 'C-' or not when calling it. So maybe there would be something better to do there.

Haxxflaxx:
I see I made some more incorrect assumptions in my first solution.
I've replaced it with explicitly getting the last key pressed in the keypad and processing it properly before setting `last-command-event`.
Also moved it out of the let so that I can set the global value.

This allows the self-insert to work like it did (probably mostly by accident) before, but also with repeat maps.

DogLooksGood:
Maybe we should patch last-command-event and last-input-event together? How do you think?

Haxxflaxx:
Yes you're right, we should also make sure that the last-input-event is correct. I also think we should wait as long as we can until we set last-command-event, and not set it when we are not executing a command.

After some thinking about it I think we could set last-input-event in `meow--keypad-handle-input-event`. We can do it the same way as I did for last-command-event in my last commit.

Then I think we should move setting last-command-event into `meow--keypad-execute`. That way it would be set for any command that would require it, as well as not setting it when we are still navigating around in the keypad.

DogLooksGood:
Why we need a undo boundary?

Haxxflaxx:
Without the boundary it won't add one when doing self-insert.
This would mean that if we do some insert, then select some text and surround it self-insert ( followed by an undo; this would result in also undoing the previous insert.

There was a boundary placed there automatically before the refactor, and this is also the behavior I observe when binding the self-insert explicitly in the keypad map.

DogLooksGood:
In beacon, we collapse the boundary so that the whole kmacro application can be reverted in one undo.

DogLooksGood:
I'm also thinking is there a way to achieve this, at the same time, remove the need on binding `H-j/H-k` stuff.  We had to bind `H-j/H-k` because in keypad leader dispatch, because the original command on `j/k` will never be found. If we can find the original command `j/k` just as how we find `self-insert-command`, then we can get rid of this tricky binding.

Haxxflaxx:
I haven't looked much into how the key rebinding to H-* is done. Are you thinking that it will work the same way as any other key to key rebinding,
e.g.
```emacs-lisp
(meow-leader-define-key
 '("L" . "H-L") ;Bind 'L' to gain access to overriden key in motion.
 '("p" . "C-x p") ;Bind 'p' for easy project menu.
 )
```
Are we doing some special magic now instead of just sending H-L?

DogLooksGood:
The way this PR works is to call a `self-insert-command` when there's no bound command found. My idea is to call the original command when there's no bound command found, therefore in most case, it will be the `self-insert-command`. It could solve this issue perfectly, at the same time, we can remove the whole idea of H-* rebind completely.

Haxxflaxx:
Ah, I see. That sounds like a good way of doing it. So an unbound key in the keypad would bypass the current meow state and call the current key map instead.

The only issue I can see is that it would be less obvious how to handle key conflicts in motion state. Currently you handle a conflict by explicitly binding something in the keypad to H-*.
This will instead solve the conflict by not binding the conflicting key in keypad.

This is mostly a problem that can be solved with documentation. It would be nice if we also had some explicit way of binding it as well, something like: `'("L" . "L")` which should just do the same as if you didn't bind it or `'("A" . "L")` if you would like to move it to another key.
I don't know if this is a great option, but it might be good to think about.

DogLooksGood:
Please try this #705 and see if it works for you.

Haxxflaxx:
I've been thinking some about this PR as well as #705, and maybe aiming for `meow-keypad-self-insert-undefined` or similar functionality may be a bit of a local maxima and a bit of an unexpected behavior.

I think what may actually be important in this PR is setting `last-input-event` and `last-command-event`.
With those a user can bind whatever they want in  the keypad, for instance all symbols to self insert for use of electric pair mode. Or throw them under a prefix like "spc s (".

@DogLooksGood what do you think about scoping out the self-insert-undefined part out of this PR, maybe even removing the option? I think removing it is what you're planning to do with #705.

DogLooksGood:
Can you try the latest #705 , I added global-map to the key lookup, now `self-insert-command` should be found as well. For me, if I set `meow-keypad-leader-transparent` to t or `normal`, `SPC (` will just insert a paren, and wrapping parentheses with region works as well.

PR #705, opened 2024-12-25
Remove motion remap

DogLooksGood:
This is an attempt to remove the rebinding in motion state.  An idea of transparent motion is introduced to achieve the same effect.

It adds a variable `meow-keypad-leader-transparent` which accepts `t`, `motion`, `normal` and `nil`. The default value is `motion` which means only in motion mode, the leader is transparent. Therefore, we no longer need the rebind to `H-*`, any free key on leader can be used to execute the original command.

NB: This will be a breaking change.


Haxxflaxx:
It works well for me in accessing overridden keys in motion state. Are we hoping for this to be able to do self insertions or will we require to explicitly bind them in the future?

Right now the meow-keypad-leader-transparent for normal state doesn't do what I would have expected, i assume we would need to have some fallback: look up the key in current-local-map, if it is not bound there, look it up in the global map instead.

I think we would still need to set last-input-event and last-command-event.

DogLooksGood:
Yep, I expect this to support `self-insert-command` as well. So it's not finished yet.

okamsn:
With these changes, would there be a built-in way to support triggering the original key binding with a separate key in the Motion state? If not, would you be willing to add a helper command that runs the binding outside of the Meow maps, such as using the code added to `meow--keypad-try-execute` in this pull request?

Then, such a helper command could be used directly, maybe in `meow--keypad-try-execute`, or wrapped and passed events or key sequences.

I think having such a command, that would need to be bound by the user, would cover the old behavior without interfering with this new proposed behavior.

DogLooksGood:
@okamsn 

I think we can add a command to achieve this.

Haxxflaxx:
I haven't really looked at how the remapping of keys works internally but I have a couple of keypad binds like `'("p" . "C-x p")`, I would expect binding something like `'("l" . "L")` to send 'L' when pressing 'l' in the keypad.
But maybe it's hard to reuse this functionality for that sort of rebinding.

Haxxflaxx:
I've done some quick testing and the transparency through to global-map works now, as well as the last-input and command -event seem to set properly.

Some things that I noticed is that when doing insert through the transparency functionality it doesn't input the undo-boundary as I would expect, but it does if I instead bind self-insert-command explicitly in the keypad.

Also it seems like meow-keypad-execute-on-beacons doesn't work for transparency. Maybe instead of calling the command directly from the `meow--keypad-try-execute` function it should call it through `meow--keypad-execute`?

DogLooksGood:
I don't think we should involve undo-boundary here. Emacs has some rules about how undo-boundary are set, the only thing Meow does about undo-boundary is in beacon state's macro application, we want to create a single undo boundary, so it's easier to cancel the whole modification.

Haxxflaxx:
A reasonable point. As I mentioned at an earlier point; it might be more reasonable to bind the self-inserting explicitly for my use case.

meow-keypad.el
```diff
@@ -439,23 +437,32 @@ try replacing the last modifier and try again."
...
+                (eq meow--keypad-previous-state meow-keypad-leader-transparent))
+          (let* ((key (meow--parse-input-event last-input-event))
+                 (local (lookup-key (current-local-map) key))
+                 (global (lookup-key (current-global-map) key))
+                 (origin-cmd (or local global))
+                 (remapped-cmd (command-remapping origin-cmd))
+                 (cmd-to-call (if (member remapped-cmd '(undefined nil))
+                                  (or origin-cmd 'undefined)
+                                remapped-cmd)))
+            (call-interactively cmd-to-call))
```
Haxxflaxx, line 451: Should this use `keypad-execute` instead?

DogLooksGood:
For your use case, you might bind `insert-pair` command.

meow-keypad.el
DogLooksGood: Yeah, I guess you are right!

capfredf:
I'm currently using Meow on the latest commit https://github.com/meow-edit/meow/commit/33414b9b1cd3207002a5b94d1300f2f4947e8126.

I assume I don't need to bind j and k to H-j and H-k using meow-leader-define-key anymore.

But I am not able to invoke what the command j or k were original bound to in a motion state buffer (e.g. a magit-status buffer) using `SPC j` or `SPC k`.  When I press `SPC`, Meow switches to the keypad state. 

<details>

<summary> Here is my meow setup. </summary>

```elisp
(defun meow-setup ()
    (setq meow-cheatsheet-layout meow-cheatsheet-layout-qwerty)
    ;; (meow-paredit-setup ())
    (meow-motion-define-key
     '("j" . meow-next)
     '("k" . meow-prev)
     '("<escape>" . ignore))
    (meow-leader-define-key
     ;; SPC j/k will run the original command in MOTION state.
     ;; '("j" . "H-j")
     ;; '("k" . "H-k")
     ;; Use SPC (0-9) for digit arguments.
     '("1" . meow-digit-argument)
     '("2" . meow-digit-argument)
     '("3" . meow-digit-argument)
     '("4" . meow-digit-argument)
     '("5" . meow-digit-argument)
     '("6" . meow-digit-argument)
     '("7" . meow-digit-argument)
     '("8" . meow-digit-argument)
     '("9" . meow-digit-argument)
     '("0" . meow-digit-argument)
     '("/" . meow-keypad-describe-key)
     '("?" . meow-cheatsheet))
    (meow-normal-define-key
     '("C-x C-f" . me/ignore-vanilla-keybindings)
     '("C-x C-b" . me/ignore-vanilla-keybindings)
     '("C-x b" . me/ignore-vanilla-keybindings)
     '("M-x" . me/ignore-vanilla-keybindings)
     '("0" . meow-expand-0)
     '("9" . meow-expand-9)
     '("8" . meow-expand-8)
     '("7" . meow-expand-7)
     '("6" . meow-expand-6)
     '("5" . meow-expand-5)
     '("4" . meow-expand-4)
     '("3" . meow-expand-3)
     '("2" . meow-expand-2)
     '("1" . meow-expand-1)
     '("-" . negative-argument)
     '(";" . meow-reverse)
     '("," . meow-inner-of-thing)
     '("." . meow-bounds-of-thing)
     '("[" . meow-beginning-of-thing)
     '("]" . meow-end-of-thing)
     '("a" . meow-append)
     '("o" . meow-open-below)
     '("b" . meow-back-word)
     '("B" . meow-back-symbol)
     '("c" . meow-change)
     '("d" . meow-kill)
     '("e" . meow-next-word)
     '("E" . meow-next-symbol)
     '("f" . meow-find)
     '("g" . meow-cancel-selection)
     '("G" . meow-grab)
     '("h" . meow-left)
     '("H" . meow-left-expand)
     '("i" . meow-insert)
     '("O" . meow-open-above)
     '("j" . meow-next)
     '("J" . meow-next-expand)
     '("k" . meow-prev)
     '("K" . meow-prev-expand)
     '("l" . meow-right)
     '("L" . meow-right-expand)
     '("m" . meow-join)
     '("n" . meow-search)
```

</details>

am I doing anything wrong here?

PR #755, opened 2025-06-28
Update meow-tutor.el

ggebbie:
Fix a super minor issue in the tutor whereby one line is made more general and correct for cases when meow-append is customized. 

Mostly this is an excuse to say THANK YOU for a great package.

### DIFF
--- a/meow-tutor.el
+++ b/meow-tutor.el
@@ -326,8 +326,8 @@
  * How to jump to the end of the current function quickly?
 
    1. Move cursor to the function below marked -->.
-   2. Type \\[meow-bounds-of-thing] and \"c\", then \"a\".
-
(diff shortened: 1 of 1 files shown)


DogLooksGood:
You are welcome. BTW, I guess that "c" can be fixed too.

ggebbie:
Good idea to update  "c". But this is not the command `meow-change` which would have been trivial to fix. This is the curly thing modifier which (I believe) is not simply a command that can be substituted into the tutorial. Similarly, the tutor lists the thing modifiers, which also would have been nice to update, but they are hard coded. Looks like the function `meow--cheatsheet-render-char-thing-table` gives an idea of how to retrieve the thing modifiers if this is promising at all. Or maybe this is not the way the tutor was envisioned in the first place.

Issue #743, opened 2025-04-04
CI: package-lint optional dependency issues (which-key)


eshrh:
CI error:
`Error: You should depend on (emacs "30.1") if you need `which-key--create-buffer-and-show'.`

which-key was added to emacs 30.

Relevant (?):
- https://github.com/meow-edit/meow/blob/e5817f3b5ef82cbcf8a729af205191dc6b142282/meow-shims.el#L389-L408
- https://github.com/purcell/package-lint/issues/281
  - purcell seems to be of the opinion that you should put each shim into a different, small package. I'm honestly inclined to agree, but it obviously involves a nonzero amount of more work.

Either way, something needs to be done about the package-lint error... recent commit (a159b743d2b3c4fa17d13e90a4c753325c320afa) was a naive attempt to quickly silence the problem but I quickly realized there's a bigger problem here. Warrants more discussion, hence this issue.

package-lint also doesn't like all the eval-after-loads in the shims but that's a different problem (or maybe not, idk)

DogLooksGood:
I have very little knowledge about CI workflow, I'm open to these tools.

jcs090218:
I've opened a PR in #761 that reproduces the same error, but for now, I've moved it into its own step with `continue-on-error: true` enabled. What do you think of this workaround? 🤔

PR #761, opened 2025-08-01
ci: Build and basic test

jcs090218:
For https://github.com/meow-edit/meow/issues/743#issuecomment-3144198937.

I've temporarily moved the `eask lint package` to its own step with `continue-on-error: true` enabled.

I've also added test coverage for both macOS and Windows.

Let me know what you think! :)

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -4,30 +4,59 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


jcs090218:
The test result is in https://github.com/jcs-PR/meow/pull/1.

PR #764, opened 2025-08-28
allow kbd-macros to also be callbacks

lanceberge:
meow forces you to bind some of the built-ins e.g. `meow-next` to keyboard macros. This makes it difficult to do rebind things like `C-n` in `meow-normal-state-keymap` because it forces you to have something bound to `next-line` in order to use `meow-next`. These changes let you additionally bind your meow kbd macros to callback functions so you can do things like:`(setq meow--kbd-forward-line #'next-line)` and then rebind `C-n`

Let me know if you'd like to keep these changes and I can document them

## Testing
to test this eval the changed buffer and eval `(setq meow--kbd-forward-line #'next-line)` then you can verify that `meow-next` works in addition to the other meow functions that use keyboard macros like `meow-prev`

### DIFF
--- a/FAQ.org
+++ b/FAQ.org
@@ -21,6 +21,14 @@ Here you can map Delete character to internal Meow's delete character function.
   (setq meow--kbd-delete-char "<deletechar>")
 #+end_src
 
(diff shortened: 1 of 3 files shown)


lanceberge:
alright no problem, I just added docs

PR #774, opened 2025-11-20
Unset insert-exit-key when it isn't needed anymore

ventruvian:
- `meow--beacon-insert-enter-key` previously stayed around
- This didn't produce any issues since it was unset by meow-beacon-start
- However cleaning it up makes it an indicator for ~~whether or not recording takes place~~. This can be helpful for user-code or other packages. _Edit: My mistake, it is not macro recording that this indicates, but execution from beacon_
- This behaviour seems more consistent otherwise it's meaning changes before the first invokcation of beacon- mode and after.
---
I resort to advicing this function to determine if ~~recording~~ macro-execution is taking place, [here](https://github.com/ventruvian/lispy-cat/blob/8a8984a0ebc142678107036984f6dd60a1c9b03f/lispy-cat.el#L55-L58) for example.
Not sure if that is a good way to do it, but regardless it seems like good style to clean this variable up.

### DIFF
--- a/meow-beacon.el
+++ b/meow-beacon.el
@@ -96,31 +96,33 @@ Non-nil BACKWARD means backward direction."
 
 (defun meow--beacon-apply-command (cmd)
   "Apply CMD in BEACON state."
-  (when meow--beacon-overlays
-    (let ((bak (overlay-get (car meow--beacon-overlays)
-                            'meow-beacon-backward)))
-      (meow--wrap-collapse-undo
-        (save-mark-and-excursion
-          (cl-loop for ov in (if bak (reverse meow--beacon-overlays) meow--beacon-overlays) do
-                   (when (and (overlayp ov))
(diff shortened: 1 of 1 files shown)


DogLooksGood:
I think there's no harm to reset it. But why not use `defining-kbd-macro` as an indicator for recording?

ventruvian:
That was the first thing I attempted, but it was `nil` in my use-case.

I realize now that I was mistaken: It is not "`recordingp`" that I need to know, but rather "`executingp`". 

I need to know if insert-functions are called from a kbd macro previously defined with beacon. I did try to switch on `executing-kbd-macro` too, but it is also `nil` at the time I need to make a decision (whether to drop into INSERT or a `lispy-mode`-enabled state).

As far as I have investigated `meow--beacon-insert-enter-key` seems to be the only state with the information I need. I'm not super happy with this either, it is certainly a very indirect way of determining that.

